PR #53, opened 2021-06-16
Experimenting with the documentation template

kdgutier:
Defining colors of the documentation.
Defining Nixtla logos.


FedericoGarza (approved):
It looks great! It has more nixtla identity.

PR #57, opened 2021-06-20
Full/Windows TSDataset split and merged ESRNN

kdgutier:
ts_dataset: greatly simplified hypars of class, full/windows split
ts_loader: minor fixes of compatibility with full/windows split
nbeats: imported new windows dataset
esrnn+mqesrnn: merged model and PytorchLightnigh wrapper
experiment_utils: fixed create_datasets class to operate with full/windows split

Pending to solve:
1) include availability and sampleability condition into ESRNN
1.1) availability condition to trim batch towards smaller series during train_step
1.2) sampleability condition to mask the gradients during training and validation
2) truncated backprop through time, large series are super slow to train
3) deprecate input_size/output_size from fullDataset


nixtlats/data/tsdataset.py
```diff
@@ -176,53 +164,33 @@ def __init__(self,
...
+# TODO: think this. porque sample_mask? no cuenta el input... ademas que pasa con random validation...
+def _define_sampleable_ts_idxs(self: BaseDataset):
+    # sum_sample_mask = self.ts_tensor[:, self.t_cols.index('sample_mask')] \
+    #                       .sum(axis=1)
+    # if self.complete_inputs:
+    #     min_mask = self.windows_size
+    # else:
+    #     min_mask = self.output_size
+    # self.samplable_ts_idxs = np.argwhere(sum_sample_mask > min_mask).reshape(1, -1)[0]
+    # self.n_samplable_ts = self.samplable_ts_idxs.size
```
FedericoGarza, line 187: It might be better to remove these lines rather than just commenting them.

nixtlats/data/tsdataset.py
```diff
@@ -312,19 +280,23 @@ def _df_to_lists(self: TimeSeriesDataset,
 
 # Cell
 @patch
-def _create_tensor(self: TimeSeriesDataset) -> None:
+def _create_tensor(self: BaseDataset) -> None:
```
FedericoGarza, line 283: This function now returns `Tuple[np.array, t.Tensor]`

nixtlats/data/tsloader.py
```diff
@@ -12,7 +12,7 @@
 from fastcore.foundation import patch
 from torch.utils.data import DataLoader
 
-from .tsdataset import TimeSeriesDataset
+from .tsdataset import TimeSeriesDataset, WindowsDataset
 
 # Cell
 class TimeSeriesLoader(DataLoader):
```
FedericoGarza, line 18: The type of the dastaset should be now `Union[TimeSeriesDataset, WindowsDataset]`.

FedericoGarza (changes requested):

nixtlats/data/tsdataset.py
kdgutier: This filtering function seems unused, I don't know if we want to preserve it.

FedericoGarza (approved):

PR #93, opened 2022-02-02
Docs

kdgutier:
Added Equations and plots to some NumPy losses.
Changed the docstrings to improve nbdev documentation.


FedericoGarza (approved):
Nice! In future iterations it would be great to change pinball loss from torch to match the rename in numpy.

kdgutier:
Sounds good, I will change the torch pinball_loss to quantile_loss

kdgutier (commented):
Added documentation and plots to NumPy losses.

FedericoGarza:
@kdgutier In the future you can use the following commands to set your credentials. With this change, each commit will have your user.

```
git config --global user.name "your_github_name"
git config --global user.email "your_github_email"
```
Also it would be great to follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).  :)

PR #156, opened 2022-02-05
Improved Long-Horizon datasets documentation

kdgutier:
- Added references and descriptions of all Long-Horizon datasets.
- Added plots of all Long-Horizon datasets.
- Reduced the example execution time.
- Pending: replication of one plot as an example.


FedericoGarza (changes requested):
I think you pushed the README.md by error. Could you revert the changes?

PR #202, opened 2022-03-20
Utils debug

kdgutier:
Fixed the bug in the call of the `model_fit_predict` function
that was creating a test Dataset with the length of the validation Dataset.


FedericoGarza (changes requested):
The changes look good, you just have to revert the changes in the README (I think it was generated by mistake).

FedericoGarza (approved):

cchallu (approved):

PR #229, opened 2022-04-14
fix: numpy smape

### LINKED ISSUE #187: SMAPELoss Numpy Implementation Error
Hello, I think there is an error in the SMAPELoss Numpy implementation. 

For the SMAPE function, the NumPy version always gives 100 times the PyTorch result.
Eg:
Numpy: 76.6044944858551
Pytorch: 0.7669

https://github.com/Nixtla/neuralforecast/blob/3248751ccdb1edab2ded9c14d9666eaddd1b1469/neuralforecast/losses/numpy.py#L218-L219

Since the description on the Numpy file was intended to map the sMAPE bound between 0% and 200%, it is reasonable to change from 200 to 2. 

https://github.com/Nixtla/neuralforecast/blob/3248751ccdb1edab2ded9c14d9666eaddd1b1469/neuralforecast/losses/numpy.py#L249

FedericoGarza:
This PR fixes #187

### DIFF
--- a/nbs/losses__numpy.ipynb
+++ b/nbs/losses__numpy.ipynb
@@ -504,9 +504,9 @@
     "    smape = 200 * np.average(smape, weights=weights, axis=axis)\n",
     "    \n",
     "    if isinstance(smape, float):\n",
-    "        assert smape <= 200, 'SMAPE should be lower than 2'\n",
+    "        assert smape <= 200, 'SMAPE should be lower than 200'\n",
     "    else:\n",
-    "        assert all(smape <= 200), 'SMAPE should be lower than 2'\n",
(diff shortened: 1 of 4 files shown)


kdgutier (changes requested):
I think it should read 'SMAPE should be lower than 200'
rather than 'SMAPE should be lower than 2'

FedericoGarza:
> I think it should read 'SMAPE should be lower than 200' rather than 'SMAPE should be lower than 2'

Fixed in [de0f3ba](https://github.com/Nixtla/neuralforecast/pull/229/commits/de0f3ba6c0dbbe28a4a28cb4365d4014f37d0122), @kdgutier

kdgutier (approved):

PR #247, opened 2022-05-11
fix reshapes and rnn

cchallu:


FedericoGarza (changes requested):
LGTM :) Could you please clean the nb outputs?🙌

PR #260, opened 2022-06-09
N-BEATS/N-HiTS predict method

kdgutier:
Extended N-BEATS and N-HiTS with the in-sample predict method that creates long format dataframe `Y_hat_df`.


neuralforecast/models/nhits/nhits.py
```diff
@@ -787,3 +787,137 @@ def forecast(self: NHITS, Y_df: pd.DataFrame, X_df: pd.DataFrame = None, S_df: p
...
+    #------------------------------------- Forecast df -------------------------------------#
+
+    # Compute per series' observations = horizon * n_windows
+    # and create unique_id column in long format
+    unique_id_counts = self.n_time_out*sample_condition.sum(dim=1) # vectorized
+    unique_ids_column = [[u_id]*count for u_id,count in zip(unique_ids, unique_id_counts)]
+    unique_ids_column = np.concatenate(unique_ids_column)
+
+    # Forecast creation date column in long format, shared across horizon
+    fcds_column = fcds.repeat(self.n_time_out)
```
FedericoGarza, line 914: These lines should be a method of the class `TimeSeriesDataset` called `get_fcds_columns` or something similar.  Particularly to avoid duplicating code in each model.

FedericoGarza (changes requested):

FedericoGarza (approved):

Issue #186, opened 2022-02-11
silent bug, training with NaNs values


FedericoGarza:
We should include an exception when missing values are passed in the `y` column. The loss becomes `nan` during training but it is hard to notice that the problem is in the data.

<img width="1067" alt="image" src="https://user-images.githubusercontent.com/10517170/153517177-c12aa313-3797-41f5-ac7b-967681c7d6b5.png">

PR #301, opened 2022-10-25
Fix normalizers

kdgutier:
Minor functionality change and SoTA TransferLearning example.
- Added available sample condition, to ensure during training that there is at least one time stamp with data.
- Added numerical protection to normalization. New normalization `z=(x-mu)/(std+eps)`.
- Added safe normalization procedure, when the `scale==0` we change it to `scale=1` to avoid exploding gradients.
- Updated the Transfer Learning notebook.
- Changed the numeration of examples towards bullets.


nbs/common.base_windows.ipynb
```diff
@@ -177,10 +177,19 @@
     "            windows = windows.reshape(-1, window_size, len(temporal_cols))\n",
```
mergenthaler, line 177: <pre class="ql-syntax" spellcheck="false">Line #186.            # print('temporal_data', temporal_data.max())
</pre><p>Are this prints necessary ? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/301/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

kdgutier: thanks for catching that

mergenthaler (approved):

mergenthaler:
@kdgutier, it would be great if we could test the improved functionality.

PR #311, opened 2022-10-31
Turned Pytorch losses into `torch.nn.module` classes

kdgutier:
Changed pytorch losses from normal classes to `torch.nn.module`, with parameters without gradient. This allows Pytorch lightning to automatically infer the gpu device where the loss parameters should be stored during training and inference.

This solves the uncertainty intervals example notebook breaking in Google Colab.


mergenthaler (changes requested):
Please include one sentence in the documentation explaining that we are inheriting `torch.nn.module` from Lighting to infer GPU devices. It might help users if something breaks.

kdgutier (commented):
You need to approve the changes

mergenthaler (approved):
Thanks for the changes.

PR #316, opened 2022-11-02
Rnns normalizers

cchallu:
- Made Scaler class agnostic to input shape, with no internal mask wrangling.
- Added dim parameter to Scaler class, allowing users (base classes) to specify the dimension to compute shift and scale. The default is -1, following the PyTorch convention.
- Added Scaler class to Base_recurrent class. The default for scaler_type is "robust".
- Modified Base_windows class to work with the latest Scaler class.
- Added tests for Scaler class.


neuralforecast/common/_base_recurrent.py
```diff
@@ -49,13 +49,10 @@ def __init__(self,
...
 
         # Scaler
-        # TODO: clean hack and add correct use of self.scaler
         if scaler_type is None:
-            self.normalize = False
             self.scaler = None
         else:
-            self.normalize = True
-            self.scaler = TemporalNorm(scaler_type=scaler_type)
+            self.scaler = TemporalNorm(scaler_type=scaler_type, dim=-1) # Time dimension is -1.
```
kdgutier, line 55: It is unclear to me that we want to set the default of `BaseRecurrent` to normalize the autoregressive channel.

We are currently using `input_size` in the `TimeSeriesDataset` to generate autoregresive channels in the last dimension and normalizing them, but a more classical approach would set `input_size=1` and apply the normalizer to the windows in dimension 1:
\# [N,C,T,L=1+H] -> [N,T,C,L=1+H] -> [N,T,C,H]

We need to rethink this.

neuralforecast/common/_base_recurrent.py
```diff
@@ -103,28 +100,45 @@ def configure_optimizers(self):
...
+        return batch
 
-    def _inv_normalization(self, y_hat, y_means, y_stds):
-        return y_stds[:, None]*y_hat + y_means[:, None]
+    def _inv_normalization(self, y_hat, temporal_cols):
+        # Receives window predictions [B, W, H]
+        # Broadcasts outputs and inverts normalization
+
+        # Get y scale and shift, and add W dimension
+        temporal_data_cols = temporal_cols.drop('available_mask')
```
kdgutier, line 132: Why did we protect `temporal_data_cols` with copy before, and here we don't?

neuralforecast/common/_base_recurrent.py
```diff
@@ -103,28 +100,45 @@ def configure_optimizers(self):
...
+        y_scale = self.scaler.x_scale[:, temporal_data_cols.get_indexer(['y']), :]
+        y_shift = self.scaler.x_shift[:, temporal_data_cols.get_indexer(['y']), :]
+
+        y_scale = torch.repeat_interleave(y_scale, repeats=y_hat.shape[1], dim=1)
+        y_shift = torch.repeat_interleave(y_shift, repeats=y_hat.shape[1], dim=1)
+
+        y_hat = self.scaler.inverse_transform(z=y_hat, x_scale=y_scale, x_shift=y_shift)
```
kdgutier, line 139: Is this transformation respecting the convention of the order for the `y_hat` dimensions?
\# [N,T,C,H,Q]?

This might be problematic.

neuralforecast/common/_scalers.py
```diff
@@ -9,48 +9,50 @@
 import torch.nn as nn
 
 # %% ../../nbs/common.scalers.ipynb 7
-def masked_median(x, mask, dim=1, keepdim=True):
+def masked_median(x, mask, dim=-1, keepdim=True):
```
kdgutier, line 12: Ibid. It is unclear that the autoregresive features channel is what we want to normalize.

kdgutier (changes requested):

neuralforecast/common/_base_recurrent.py
cchallu: yes

neuralforecast/common/_base_recurrent.py
cchallu: because we drop available_mask

kdgutier (approved):

neuralforecast/common/_scalers.py
cchallu: This is only the default, set to -1 to follow PyTorch conventions. Different base classes have the time in different dimensions of the tensor

PR #320, opened 2022-11-04
rnns with decoders, autos, usage examples, fix val

cchallu:
- Improved recurrent models (RNN, LSTM, GRU, DilatedRNN) decoder with MLP and contexts.
- Added truncated backpropagation to recurrent models, controlled with input_size hyperparameter.
- RNNs input only includes information on the current timestamp (no lags).
- Fixed bug in validation_step in base_recurrent class.
- Updated default configuration of auto models to reflect changes in architecture.
- Improved images and documentation of recurrent models.
- Improved docstring.


nbs/common.modules.ipynb
```diff
@@ -24,7 +24,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Encoders"
+    "# PyTorch Modules"
```
kdgutier, line 27: Can we change title to NeuralForecast Modules?

kdgutier (approved):

PR #347, opened 2022-12-01
Feature: added return_params feature to distribution classes

dluuo:
Changes:
Added return_params argument to PMM/GMM/DistributionLoss pytorch classes
Changed predict_step in base_windows/base_recurrent to return params if return_params flag = True


kdgutier:
Thanks for the PR.
I changed the essence of the changes to https://github.com/Nixtla/neuralforecast/pull/348.
Current branch has some merge conflicts.

PR #369, opened 2022-12-14
[FIX] Losses in Auto models initialization

cchallu:
Fixed a bug with the default config for Auto models caused by defining multiple losses.

- Losses for Auto models can only be defined in the initialization, defaulted with MAE.
- Added asserts in BaseAuto to avoid passing horizon and loss inside the configs.
- deepcopy the config input inside BaseAuto to avoid overwriting it (with horizon and loss).


neuralforecast/common/_base_auto.py
```diff
@@ -112,10 +113,23 @@ def __init__(
     ):
         super(BaseAuto, self).__init__()
         self.save_hyperparameters()  # Allows instantiation from a checkpoint from class
-        config["h"] = h
+
+        assert (
```
kdgutier, line 117: I think that it is convenient to change these asserts for if conditions with Exception messages.
The message helps user's debugging.
```
        if not self.fitted:
            raise Exception("This model instance is not fitted yet, Call fit method.")
```

kdgutier (approved):

PR #374, opened 2022-12-16
[DOCS] Updated tutoriales with latest pip version

cchallu:
* Updated Long Horizon examples to be compatible with the latest pip version. Moved loss from config to model init.
* Updated Probabilistic example, moved loss from config to model init.
* Added more details to Save and load tutorial.
* Minor edit in Installation guide.


mergenthaler (approved):
LGTM

PR #397, opened 2023-01-10
[FIX] Return scaled distribution parameters

cchallu:
* Parameters of PMM and GMM are now returned by `predict` in the original scale.
* Added `get_params` method to distribution losses which returns parameters in original scale.


neuralforecast/losses/pytorch.py
```diff
@@ -798,6 +798,28 @@ def get_distribution(
                 distr, [AffineTransform(loc=loc, scale=scale)]
             )
 
+    def get_params(
```
kdgutier, line 801: Can we add an issue with the pending work of homogenization of the `get_params` and the `get_distribution` methods?

kdgutier (approved):

PR #414, opened 2023-01-23
Fix : Updated Scale Decouple Optimization (for Variance)

dluuo:
Reverted Variance Scale Decouple Optimization for GMM class, and Distribution Loss class StudentT/Normal.

Current Scale Decouple:
```
eps = (eps * scale)
stds = F.softplus(stds) + eps
```

In this PR:
```
stds = F.softplus(stds)
stds = (stds + eps) * scale
```

We empirically tested the scale_decouple versions on hierarchical `Traffic, Tourism, TourismLarge, Wiki, Labour` datasets, and found that with carefully tuned learning rates the second version improves on the first.


nbs/losses.pytorch.ipynb
```diff
@@ -1127,12 +1127,13 @@
     "    variance and residual location based on anchoring `loc`, `scale`.\n",
     "    Also adds StudentT domain protection to the distribution parameters.\n",
     "    \"\"\"\n",
+    "    print(\"new student scale decoouple\")\n",
     "    df, mean, tscale = output\n",
+    "    tscale = F.softplus(tscale)\n",
```
kdgutier, line 1132: Can we take this print out?

kdgutier (approved):

nbs/losses.pytorch.ipynb
dluuo: resolved

kdgutier (approved):

PR #424, opened 2023-01-28
[FEAT] Rich Fix + Negative Binomial

kdgutier:
- Eliminating `rich` dependency from `settings.ini` to fix broken Colab environment.
- Improving Negative Binomial scale decouple.


FedericoGarza (approved):
LGTM

PR #425, opened 2023-01-30
[FIX] Fixed NaNs in scalers with short series

cchallu:
Fixed a bug when scalers return NaNs when time series are short.

Protect masked_mean and masked_median to return zeros when all values are masked.


kdgutier (approved):
Is this scaler problem restricted only to `nan_median` in the `robust` version.
What happens with `standard` version?

FedericoGarza:
Nice! It might be great to have a unit test to ensure the change works as expected.

PR #429, opened 2023-02-02
[FEAT] Added predict_rolled method

cchallu:
New `predict_rolled` method in `core` class to predict historical values on training data.


neuralforecast/core.py
```diff
@@ -319,6 +320,8 @@ def cross_validation(
             Length of test size. If passed, set `n_windows=None`.
         sort_df : bool (default=True)
             Sort `df` before fitting.
+        fit_models: bool (default=True)
```
kdgutier, line 323: I think that the description of the arguments function can improve.
to `bool=False, Whether or not we fit models during cross-validation.`

neuralforecast/core.py
```diff
@@ -400,6 +409,60 @@ def cross_validation(
...
+        n_windows: int = 1,
+        step_size: int = 1,
+        insample_size: Optional[int] = None,
+        sort_df: bool = True,
+        verbose: bool = False,
+        **data_kwargs,
+    ):
+        """Predict insample with core.NeuralForecast.
+
+        Use stored fitted `models` to predict historic values of a time series from DataFrame `df`.
```
kdgutier, line 425: Can we add a comment on possibility of refit during cross-validation?

kdgutier (approved):

kdgutier:
Let's change the name of the new method from `predict_in_sample` to `rolled_predictions` to not generate false expectations on the filtered train/in-sample predictions. The test filters are pending.

PR #447, opened 2023-02-15
[FEAT] Add gpu tests

FedericoGarza:
- feat: add gpu config pipeline
- feat: add gpu pipeline test


FedericoGarza:
we need circle ci approval to use gpu

PR #450, opened 2023-02-20
[DOCS] Electricity Peak Detection

cchallu:
Tutorial with using the `NHITS` model for peak detection on the ERCOT electricity market.


FedericoGarza (approved):
LGTM

PR #463, opened 2023-03-09
[FEAT] Informer, augment_calendar_df, set seeds in fit and predict

cchallu:
1. New univariate Informer model.
2. New AutoInformer model.
3. Added Transformers modules to `common_modules`.
4. Added Informer in `core` tests and CI performance tests.
5. Added `augment_calendar_df` function in `utils` based on Informer's calendar feature generation.
6. Restart seed in `fit` and `predict` methods to control randomness from sampling (distribution losses and forwards of some models)


nbs/core.ipynb
```diff
@@ -67,7 +67,7 @@
     "from neuralforecast.models import (\n",
     "    GRU, LSTM, RNN, TCN, DilatedRNN,\n",
```
kdgutier, line 68: Would we want to add a unit test of `fit`, `predict` new random seed argument.
To guarantee replicability of their expected results in the future?

nbs/utils.ipynb
```diff
@@ -366,12 +366,173 @@
     "ax.grid()"
```
kdgutier, line 366: Is it possible to use show_doc on the new functions for the calendar features documentation?

neuralforecast/utils.py
```diff
@@ -289,3 +291,155 @@ def generate_series(
...
+                HourOfDay,
+                DayOfWeek,
+                DayOfMonth,
+                DayOfYear,
+            ]
+        ]
+
+
+def augment_calendar_df(df, freq="h"):
+    """
```
kdgutier, line 408: Can we standardize a bit this docstring and show it in `utils.ipynb`?

kdgutier (approved):

PR #499, opened 2023-04-04
HINT Unit Test

dluuo:
Added a Unit Test for the HINT notebook.
It generates a random dataset, runs the HINT network on the dataset, and then checks that the outputs satisfy hierarchical coherence (within a threshold epsilon)
Also fixed minor typo in HINT class


dluuo:
Note sure why core.ipynb test is failing, since I do not change it in this PR

PR #503, opened 2023-04-05
[FIX] Prevent using pretrained weights

FedericoGarza:
The `fit` and `cross_validation` methods currently update the models' weights in each execution. This can result in unexpected results, especially if the same class is used on different datasets (the weights would be "transferred"). Additionally, this behavior has the limitation that multiple sequential executions of the `fit` method do not retrieve the same forecasts (i.e., after the first `fit`, the following ones will retrain the model with the already adjusted weights, which can cause confusion regarding the training step). This PR `deepcopy` the models before training them to avoid this unexpected behavior. Several tests for `fit`, `cross_validation`, and their combination were included, particularly the following one that used to fail:

```
models = [NHITS(h=12, input_size=24, max_steps=10)]
nf = NeuralForecast(models=models, freq='M')
nf.fit(AirPassengersPanel_train)
init_fcst = nf.predict()
nf.fit(AirPassengersPanel_train)
after_fcst = nf.predict()
test_eq(init_fcst, after_fcst)
```


cchallu:
I am not sure we want to have this function as default. It is expected that fitting a model will change the weights of the model. It is also a nice capability to have pre-trained models and fine-tune them. Users can re-instantiate the model again if they want to recover the original model (the seed will ensure the same initialization). At most, I suggest having a flag to have the deep-copy, but have False as default.

PR #517, opened 2023-04-12
removed duplicate from model collection list

vinish-ai:
Just removed the duplicate model name(Informer) from the model collection list


kdgutier:
Hey @VinishUchiha, 

Thanks for catching the duplicated `Informer`.

PR #540, opened 2023-04-22
[FIX] Predict dates bug

cchallu:
- Fixed a bug on the predict function when dates and uids in `df` do not match the stored dataset passed during `train` or `cross_validation`.
- Replaced `_make_future_df` with `_future_dates` to homologize with other date creating functions.


FedericoGarza (approved):
Cool! LGTM :)

PR #554, opened 2023-05-02
[FEAT] Adding robust point forecasting `HuberLoss`

hugogobato:
SmoothL1Loss creates a criterion that uses a squared term if the absolute element-wise error falls below beta and an L1 term otherwise. It is less sensitive to outliers than MSE and in some cases prevents exploding gradients (e.g. see the paper [Fast R-CNN](https://arxiv.org/abs/1504.08083) by Ross Girshick).

I added at the bottom of the file the code for this loss function and to do so, I needed to also add "import torch.nn as nn" in the beginning of the file.

### DIFF
--- a/neuralforecast/losses/pytorch.py
+++ b/neuralforecast/losses/pytorch.py
@@ -2,7 +2,7 @@
 
 # %% auto 0
 __all__ = ['MAE', 'MSE', 'RMSE', 'MAPE', 'SMAPE', 'MASE', 'QuantileLoss', 'Accuracy', 'MQLoss', 'wMQLoss', 'sCRPS', 'relMSE',
-           'DistributionLoss', 'PMM', 'GMM', 'NBMM']
+           'DistributionLoss', 'PMM', 'GMM', 'NBMM', 'SmoothL1Loss']
 
 # %% ../../nbs/losses.pytorch.ipynb 3
 from typing import Optional, Union, Tuple
@@ -11,6 +11,7 @@ import math
 import numpy as np
 import torch
 
+import torch.nn as nn
 import torch.nn.functional as F
 from torch.distributions import Distribution
 from torch.distributions import Bernoulli, Normal, StudentT, Poisson, NegativeBinomial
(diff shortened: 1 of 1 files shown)


kdgutier:
Hey @hugogobato,

Thanks for the contribution to the NeuralForecast library.
To create the library we are using `nbdev`, here are the [Contribution instructions](https://github.com/Nixtla/neuralforecast/blob/main/CONTRIBUTING.md).

Additional comments on the PR:
- Let's rename the class into `HuberLoss` after the Peter Huber, that originally proposed it.
- Let's update the docstring into:
    ```
    """ Huber Loss
    Creates a criterion that uses a squared term if the absolute element-wise error falls below beta and an L1 term otherwise. 
    It is less sensitive to outliers than MSE and in some cases prevents exploding gradients
    $$ l_n = \begin{cases}
    0.5 (x_n - y_n)^2,   \text{        if } |x_n - y_n| < \delta \\
    \\delta * (|x_n - y_n| - 0.5 * \\delta),   \text{     otherwise }\end{cases}$$

    **Parameters:**
    `delta`: float=0.5. The slope of the quantile loss, in the context of quantile regression, the q determines the conditional quantile level.

    **References:**
    [Huber Peter, J (1964). "Robust Estimation of a Location Parameter". Annals of Statistics](https://projecteuclid.org/journals/annals-of-mathematical-statistics/volume-35/issue-1/Robust-Estimation-of-a-Location-Parameter/10.1214/aoms/1177703732.full)
    """
    ```
- To add the `HuberLoss` you would need to change this notebook https://github.com/Nixtla/neuralforecast/blob/main/nbs/losses.pytorch.ipynb.
- After adding the HuberLoss don't forget to add the documentation in the notebook:
  ```
  show_doc(HuberLoss, name='HuberLoss.__init__', title_level=3)
  show_doc(HuberLoss.__call__, name='HuberLoss.__call__', title_level=3)
  ``` 
- Let's change the `HuberLoss`' location into group A, the scale dependent errors collection to maintain the losses taxonomy.

PR #556, opened 2023-05-02
[Fix] AutoNHITS search space types and improved documentation

kdgutier:
- Improved `NHITS`' `stack_types` parameter documentation, to not include trend and seasonality.
- Changed `n_pool_kernel` to include `NHITS`' default.
- Changed `max_steps` to be sampled under smooth assumptions with `tune.quniform`.
- Changed `batch_size` to be sampled under smooth assumptions with `tune.qloguniform`.
- Changed `windows_batch_size` to be sampled under smooth assumptions with `tune.qloguniform`.
- Changed `random_seed` to be sampled under independent assumptions with `tune.randint`.


candalfigomoro:
@kdgutier 

Also `n_freq_downsample` does not include the default `[4, 2, 1]` value (see https://github.com/Nixtla/neuralforecast/blob/main/neuralforecast/models/nhits.py#L237).

Also, the same improvements could also be applied to the other Auto* classes ('AutoRNN', 'AutoLSTM', 'AutoGRU', 'AutoTCN', 'AutoNBEATS', 'AutoNBEATSx' etc.)

candalfigomoro:
@kdgutier 

With the new code, I also get errors such as:

```
ValueError: batch_size should be a positive integer value, but got batch_size=8.0
```

Something like this seems to work:

```
    "batch_size": tune.sample_from(
        lambda _: int(2 ** np.random.randint(np.log2(32), np.log2(256) + 1))
    ),
    "windows_batch_size": tune.sample_from(
        lambda _: int(2 ** np.random.randint(np.log2(128), np.log2(1024) + 1))
    ),
```

But maybe there's a cleaner solution.

candalfigomoro:
@kdgutier You're right, using np.random.randint makes no sense. Looking at the Ray documentation https://docs.ray.io/en/latest/tune/api/search_space.html it seems that we can use another value from the search space inside `tune.sample_from()` (instead of using np.random), so that could be another option.

cchallu (changes requested):
modify tutorials to reflect new changes

cchallu (approved):

PR #576, opened 2023-05-17
fix freq default for augment_calendar_df

tylernisonoff:
`h` needs to capitalized


kdgutier:
Hey @tylernisonoff,

Thanks for the fix, in order to include it in the library you would need to execute it through `nbdev`.
Here are the contributing guidelines: https://github.com/Nixtla/neuralforecast/blob/main/CONTRIBUTING.md

If you have troubles with `nbdev`, ping me.

tylernisonoff:
@kdgutier  my apologies! Should have looked harder for the contributing doc 😅 

I believe I've made the right changes -- however a modification to `neuralforecast/_modidx.py` was included from running `nbdev_export`

I've left it here so you can take a look if you'd like, but happy to remove from the commit / PR if it should not be included.

kdgutier (approved):

kdgutier:
Pretty cool @tylernisonoff, I am running the tests.
Thanks for the `augment_calendar_df` default's fix.

PR #595, opened 2023-05-25
[Feat] NBEATSx dropout

kdgutier:
Added dropout to NBEATSx.
Since individual models benefit from dropout regularization, this is the most common use case.


cchallu (approved):

andrewrgarcia:
Doesn't NBEATSx also rely on ensembling? I thought dropout lowered the performance of ensemble

PR #598, opened 2023-05-26
[FEAT] Val_check_step protection

cchallu:


kdgutier (approved):

kdgutier (commented):
It is convenient to have a tutorial or explanation to make the transition between max_epochs and max_steps smooth.

Basically telling people to use:
```
# WindowsBased
max_steps = len(Y_train_df) // batch_size

# RecurrentBased
max_steps = len(Y_train_df.unique_id.unique()) // batch_size
```

PR #618, opened 2023-06-07
[FEAT]  Train/Val NaN Loss protection

cchallu:
* Stop training after train or validation loss is NaN, prints NaNs in inputs and outputs to debug.
* Removed several models from CI evaluation performance to speed up tests.


FedericoGarza (approved):

kdgutier (commented):
cool

PR #661, opened 2023-06-14
[DOCS] Neuralforecast map and How-to add new models.

cchallu:
Two new tutorials:

* NeuralForecast map
* How to add new models


FedericoGarza (approved):
Pretty cool, @cchallu! Overall the tutorials reflect really well the main characteristics of the library and are well-written. Particularly, I liked that you highlighted the advantages of adding a new model to NeuralForecast and the table describing the shape of the tensors for the different base classes. It would be really helpful even for maintainers. 😍

I have a couple of comments that could improve the tutorials, but they are non-blocking:

- A note that the map refers to the different methods of particular classes. For example,  we could include `NeuralForecast` instead of `Core` or `Core::Neuralforecast` to improve readability.
- An additional step on how to add the new model to the CI tests.

PR #668, opened 2023-06-15
[Fix] Hint missing early_stopping_steps

kdgutier:
- A new core protection was keeping HINT from running correctly.


nbs/common.base_windows.ipynb
cchallu, line 1: Please remove this change to prioritize the previous PR.

neuralforecast/common/_base_windows.py
cchallu, line 1: Idem above

cchallu (changes requested):

cchallu (approved):

PR #670, opened 2023-06-19
Feature/insample test

yguooo:
Implement the custom dataset input for `predict_insample` function. It allows evaluating the same model but different dataset from the most recently used dataset. For example, I want evaluate `predict_insample` on test data given a model trained in train data. Changes mostly involve simple modifications of the dataset variables.


marcopeix:
Closing. Duplicate of #940

PR #682, opened 2023-07-01
[FEAT] Added DeepAR model 

kdgutier:
First, commit to the DeepAR model. It inspires by TFT's Seq2Seq approach and simplifies it to DeepAR's original architecture.

Some things to consider in future work on DeepAR:
- Rethink the usage of Variable Selection Networks; we might want to use simpler encoders closer to GluonTS'.
- Rethink the usage of TFT embeddings; same as with VSN's we might want to use simpler encoders closer to GluonTS'.


cchallu:
Missing recursive forecasting method, which in my opinion is the main difference between DeepAR an regular LSTM-based methods we already have.

kdgutier:
> Missing recursive forecasting method, which in my opinion is the main difference between DeepAR an regular LSTM-based methods we already have.

This is not the case. 

In its current version, DeepAR is recursive because it uses a decoder RNN. The main difference to the original version is that we are not relying on Monte Carlo sampling that feeds into the autoregressive features.

cchallu:
This is what I meant, the recursive forecasting strategy as opposed to direct.

PR #690, opened 2023-07-12
[FIX] Change default of `use_init_models`

cchallu:
Change the default value of `use_init_models` to False to fine-tune as the default behavior.

Print warning when the previously fitted model is deleted.

### DIFF
--- a/neuralforecast/core.py
+++ b/neuralforecast/core.py
@@ -194,7 +194,7 @@ class NeuralForecast:
         static_df: Optional[pd.DataFrame] = None,
         val_size: Optional[int] = 0,
         sort_df: bool = True,
-        use_init_models: bool = True,
+        use_init_models: bool = False,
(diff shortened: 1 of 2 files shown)


neuralforecast/core.py
```diff
@@ -213,7 +213,7 @@ def fit(
             Size of validation set.
         sort_df : bool, optional (default=False)
             Sort `df` before fitting.
-        use_init_models : bool, optional (default=True)
+        use_init_models : bool, optional (default=False)
             Use initial model passed when NeuralForecast object was instantiated.
```
kdgutier, line 217: Can you add the behavior when use_init_models = False?
```
If False, fit is performed on pre-fitted models.
```

kdgutier (approved):

FedericoGarza (approved):
LGTM! :)

PR #720, opened 2023-08-15
[FIX] Fixed bug in validation loss scale

cchallu:
Fixed a bug in the computation of the validation loss for windows-based models when scaling and using point losses. The validation loss was comparing scaled predictions against real values in the original scale. This impacted both early stopping and model selection for auto models.


FedericoGarza (approved):
Nice @cchallu! LGTM :)

Issue #526, opened 2023-04-16
[Auto] FileNotFoundError on Windows


soutrik7771:
# Cuurently I am showcasing the error with the example fgiven in the doc, facing the same issue with my office data which is in exact same format as expected by the module

Y_df = pd.read_parquet('https://datasets-nixtla.s3.amazonaws.com/m4-hourly.parquet')
# subsizing for quick experiment
max_len = 10 * 24
Y_df = Y_df.groupby("unique_id").tail(
    10 * 24
)  # Select last 10 days of data to make example faster
# ensure ds ends at the same point, avoid errors in cv
Y_df["ds"] = Y_df.groupby("unique_id")["ds"].transform(
    lambda x: range(len(x) - max_len + 1, max_len + 1)
)

nf = NeuralForecast(
    models=[
        AutoNHITS(
        h=ft_unit,
            loss=SMAPE(),
            num_samples=11,
            cpus=int(os.cpu_count() - 1),
            verbose=True,
        ),
    ],
    freq="H",
)

nf.fit(df=Y_df)

### Getting this chain of error from every thread opened during parallel processing---

(train_tune pid=15316) Global seed set to 4
2023-04-16 06:04:30,676	ERROR trial_runner.py:1062 -- Trial train_tune_0342e_00000: Error processing event.
ray.exceptions.RayTaskError(NotImplementedError): ray::ImplicitFunc.train() (pid=15316, ip=10.231.1.6, repr=train_tune)
  File "/anaconda/envs/pnl_py37/lib/python3.9/site-packages/ray/tune/trainable/trainable.py", line 368, in train
    raise skipped from exception_cause(skipped)
  File "/anaconda/envs/pnl_py37/lib/python3.9/site-packages/ray/tune/trainable/function_trainable.py", line 337, in entrypoint
    return self._trainable_func(
  File "/anaconda/envs/pnl_py37/lib/python3.9/site-packages/ray/tune/trainable/function_trainable.py", line 654, in _trainable_func
    output = fn()
  File "/anaconda/envs/pnl_py37/lib/python3.9/site-packages/ray/tune/trainable/util.py", line 406, in _inner
    return inner(config, checkpoint_dir=None)
  File "/anaconda/envs/pnl_py37/lib/python3.9/site-packages/ray/tune/trainable/util.py", line 398, in inner
    return trainable(config, **fn_kwargs)
[...]

PR #735, opened 2023-08-28
On Windows, prevent long trial directory names

tg2k:
On Windows, prevent long trial directory names (https://github.com/ray-project/ray/issues/29586, https://github.com/Nixtla/neuralforecast/issues/526).


tg2k:
Thanks @jmoralez, I hadn't realized. I pushed another change that may help.

jmoralez:
Hey @tg2k. Are you able to provide a reproducible example of this? I've been trying to do that in #814 but it passes haha. If you can provide one I can include it there and merge this to make sure it fixes the issue. Then we can merge this to main.

tg2k:
@jmoralez It's been about 3 months since I made a stab at a Windows environment, so my memory in this is a little fuzzy, and I don't have a reproducible example.

I believe my Windows-side directory path was 21 characters long. As an example of a Windows path that could be created, from https://github.com/Nixtla/neuralforecast/issues/526#issuecomment-1515829035 you can see a path like 

`C:\pathto\ray_results\train_tune_2023-04-20_15-07-40\train_tune_7f3f4_00000_0_check_val_every_n_epoch=100,hidden_size=512,hist_exog_list=spx_log_ret,input_size=20,learning_rate=0.0012_2023-04-20_15-07-43\lightning_logs\version_0\events.out.tfevents.1681974471.Desktop-DC.30528.0`

which is about 279 characters long. Windows maxes out at [260 characters](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry) unless you enable extended-length paths (as shown on that page), but for the most part support for those long paths is weak in Windows.

You can see from the example above that the default behavior encodes a lot of parameters into the path. If you have a long enough base path, or enough parameters (such as exogenous variables), you'll push the path over the limit (unless Ray-Tune has changed how it creates paths on Windows in the meantime).

elephaint:
Trying to reproduce on Windows 11, no luck yet. 

A potential issue with this fix is that `trial_dirname_creator` is an [alpha feature](https://docs.ray.io/en/latest/tune/api/doc/ray.tune.TuneConfig.html)

jmoralez (approved):

PR #763, opened 2023-09-27
support optuna in auto models

jmoralez:
Adds the possibility of using [optuna](https://optuna.org/) to search for hyperparameters for the Auto* models by setting `backend='optuna'` and passing a function to the `config` argument that takes a trial as input and returns a dictionary with a combination of hyperparameters.


settings.ini
cchallu, line 1: We should also add optuna to requirements (not only dev).

cchallu (approved):

PR #801, opened 2023-10-30
Fix Crossvalidation when dataframe is not provided

### LINKED ISSUE #800: [Core] CV fails when df is not provided

This line https://github.com/Nixtla/neuralforecast/blob/ff1566c2af4c76284e8f432b4ae82f6dfffd14e6/neuralforecast/core.py#L542
tries to merge with `df`, which can be `None`, so in that case it fails.

Originally reported [here](https://nixtlacommunity.slack.com/archives/C031M8RLC66/p1698432499450449)

### Versions / Dependencies

1.6.4

### Reproduction script

```
model = NeuralForecast(
                [AutoNBEATS(**model_args)],
                freq='D',
            )
model.fit(df=data, val_size=horizon)
crossvalidation_df = model.cross_validation()
```

### Issue Severity

americast:
The error arising when a dataframe is not passed when `cross_validation` is called is addressed in this PR. Fixes #800.

Looking forward to reviews and suggestions.


marcopeix:
#1545 fixes the issue.

PR #833, opened 2023-11-30
add frequency validation and futr_df debugging methods

jmoralez:
Adds a frequency validation that makes sure the times and frequency match (e.g. both ints, timestamps + pandas or polars offset).
Also adds the following methods:
* `NeuralForecast.make_future_dataframe`: provides the expected combinations of ids and dates to serve as a base when providing `futr_df`.
* `NeuralForecast.get_missing_future`: takes the `futr_df` and returns which combinations are missing for easier debugging.


cchallu (approved):
Awesome features :)

PR #842, opened 2023-12-14
retrain cv

jmoralez:


nbs/core.ipynb
```diff
@@ -292,7 +292,6 @@
     "\n",
```
cchallu, line 292: <pre class="ql-syntax" spellcheck="false">Line #572.            splits = ufp.backtest_splits(
</pre><p>I think it will raise error if refit=True and n_windows=None and user only specifies test_size. We might want to move the conversion n_windows&lt;&gt;test_size in _no_refit_cross_validation (line 422 approx) to get n_windows from test_size.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/842/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

cchallu, line 292: <pre class="ql-syntax" spellcheck="false">Line #583.            for i_window, (cutoffs, train, test) in enumerate(splits):
</pre><p> </p><p>This is a comment for future us. I foresee that pre-computing and keeping in memory all splits can cause memory issues if datasets are large and it has many splits, since dataset is almost multiplied number of splits times. Is there any advantage on this approach vs creating and keeping in memory only the current split? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/842/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

cchallu, line 292: <pre class="ql-syntax" spellcheck="false">Line #599.                    futr_df: Optional[DataFrame] = test
</pre><p>Remove Optinal[DataFrame]?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/842/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez: <p>This is in case we adopt mypy, if we ran it it would complain about futr_df being declared as DataFrame and then assigned as None in the else below.</p>

jmoralez: The backtest_splits [returns a generator](https://github.com/Nixtla/utilsforecast/blob/main/utilsforecast/processing.py#L740), so this won't be an issue.

jmoralez: addressed in 96e87d9

cchallu (approved):

PR #849, opened 2024-01-02
use future instead of deprecation warnings

jmoralez:
The default warning filters ignore deprecation warnings ([docs](https://docs.python.org/3/library/warnings.html#default-warning-filter)). This changes the category of the issued warnings from `DeprecationWarning` to `FutureWarning` in order to provide more visibility of future breaking changes to our users.


cchallu (approved):
Awesome!

PR #862, opened 2024-01-17
[FIX] Polars `.columns` produces list rather than Pandas Index

akmalsoliev:
Fixes the handling of `static_df` when it's a polars DataFrame.


neuralforecast/tsdataset.py
```diff
@@ -317,8 +317,10 @@ def from_df(
 
         # Static features
         if static_df is not None:
-            static_cols = static_df.columns.drop(id_col)
+            static_cols = static_df.drop(columns=id_col).columns
```
jmoralez, line 320:
```suggestion
            static_cols = pd.Index([c for c in static_df.columns if c != id_col])
```

neuralforecast/tsdataset.py
```diff
@@ -317,8 +317,10 @@ def from_df(
...
             static = ufp.to_numpy(static_df[static_cols])
+            if not isinstance(static_cols, pd.Index):
+                static_cols = pd.Index(static_cols)
```
jmoralez, line 323:
```suggestion
```

jmoralez (changes requested):
Thanks for the fix. Just a small suggestion

jmoralez (approved):
Thanks!

PR #871, opened 2024-01-23
switch `flake8` to `ruff`

### LINKED ISSUE #868: switch to Ruff instead of `flake8`
### Description

Recently [Ruff](https://github.com/astral-sh/ruff) has become very popular among also mainstream projects for:
- much faster, they say 70x
- integrate several other tools under one roof, such as isort or bandit

### Use case

if you are fine with this change I am happy to send such PR for review... :rabbit:

Borda:
resolves #868

cc: @jmoralez

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.ruff]
 target-version = "py38"
-line-length = 120
+line-length = 88
 # Enable Pyflakes `E` and `F` codes by default.
 select = [
 #    "E",
@@ -14,4 +14,4 @@ select = [
 
 
 [tool.mypy]
-ignore_missing_imports = "True"
+ignore_missing_imports = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,17 @@
+[tool.ruff]
+target-version = "py38"
+line-length = 120
+# Enable Pyflakes `E` and `F` codes by default.
+select = [
+#    "E",
+#    "W", # see: https://pypi.org/project/pycodestyle
+    "F", # see: https://pypi.org/project/pyflakes
+#    "I", #see: https://pypi.org/project/isort/
+#    "D", # see: https://pypi.org/project/pydocstyle
+#    "N", # see: https://pypi.org/project/pep8-naming
+#    "S", # see: https://pypi.org/project/flake8-bandit
+]
+
+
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -0,0 +1,17 @@
+[tool.ruff]
+target-version = "py38"
+line-length = 120
+# Enable Pyflakes `E` and `F` codes by default.
+select = [
+#    "E",
+#    "W", # see: https://pypi.org/project/pycodestyle
```
Borda, line 7: In the following PRs, I would suggest applying a few more basic checks

pyproject.toml
```diff
@@ -0,0 +1,17 @@
+[tool.ruff]
+target-version = "py38"
+line-length = 120
```
jmoralez, line 3: We use black's default of 88
```suggestion
line-length = 88
```

pyproject.toml
```diff
@@ -0,0 +1,17 @@
...
+    "F", # see: https://pypi.org/project/pyflakes
+#    "I", #see: https://pypi.org/project/isort/
+#    "D", # see: https://pypi.org/project/pydocstyle
+#    "N", # see: https://pypi.org/project/pep8-naming
+#    "S", # see: https://pypi.org/project/flake8-bandit
+]
+
+
+[tool.mypy]
+ignore_missing_imports = "True"
```
jmoralez, line 17: Not sure if it makes a difference
```suggestion
ignore_missing_imports = true
```

jmoralez (changes requested):

pyproject.toml
Borda: That may not be valid yaml syntax

pyproject.toml
Borda: Right, but as you check only for `F` and no `E` issues, long lines are skipped already :)
So this change has no effect and to be clear, it did not also have now...
just with enabling check for long lines, there are a few tens or hundreds issue :rabbit:

pyproject.toml
jmoralez: But it's TOML. You can see it should be lowercase [here](https://mypy.readthedocs.io/en/stable/config_file.html#example-pyproject-toml).

jmoralez (approved):

PR #880, opened 2024-01-26
[FEAT] Model table in README

cchallu:
Added model's table in readme


jmoralez (approved):

candalfigomoro:
@cchallu Maybe the green check for Exogenous features with Informer, Autoformer and FEDformer is a little bit "optimistic" given that they don't support historical and static exogenous variables?

PR #891, opened 2024-02-15
Feat/nf evaluation

marcopeix:
Script to evaluate all NF models on benchmark datasets. The table needs to be updated with metrics as results are obtained. 

Main changes:

- Remove "alias" in models.py and simply use the model name instead
- Specify the `ds` column as `int` or `datetime` in datasets.py
- Add README and environment.yml file for reproducibility and to show results


experiments/nf_evaluation/datasets.py
```diff
@@ -0,0 +1,77 @@
+import pandas as pd
+
+from datasetsforecast.m4 import M4
+from datasetsforecast.m3 import M3
+from datasetsforecast.long_horizon import LongHorizon
+
+
+def get_dataset(name):
```
VascoSch92, line 8: the name of the method is not correct. You are returning only the dataset but also other parameters

experiments/nf_evaluation/models.py
```diff
@@ -0,0 +1,30 @@
+from neuralforecast.auto import *
```
VascoSch92, line 1: I will import just what you need and not everything

experiments/nf_evaluation/models.py
```diff
@@ -0,0 +1,30 @@
...
+from neuralforecast.losses.pytorch import HuberLoss, DistributionLoss
+
+
+def get_model(model_name, horizon, num_samples):
+    """Returns the model class given the model name.
+    """
```
VascoSch92, line 7: nit: docstring on one line, i.e., `"""Returns the model class given the model name."""`

experiments/nf_evaluation/models.py
```diff
@@ -0,0 +1,30 @@
...
+        'AutoTFT': AutoTFT(config=None, h=horizon, loss=HuberLoss(), num_samples=num_samples),
+        'AutoVanillaTransformer': AutoVanillaTransformer(config=None, h=horizon, loss=HuberLoss(), num_samples=num_samples),
+        'AutoInformer': AutoInformer(config=None, h=horizon, loss=HuberLoss(), num_samples=num_samples),
+        'AutoAutoformer': AutoAutoformer(config=None, h=horizon, loss=HuberLoss(), num_samples=num_samples),
+        'AutoFEDformer': AutoFEDformer(config=None, h=horizon, loss=HuberLoss(), num_samples=num_samples),
+        'AutoTimesNet': AutoTimesNet(config=None, h=horizon, loss=HuberLoss(), num_samples=num_samples),
+        'AutoPatchTST': AutoPatchTST(config=None, h=horizon, loss=HuberLoss(), num_samples=num_samples)
+    }
+
+    return model_dict[model_name]
```
VascoSch92, line 30: you could just create a dictionary and import it instead of create a dictionary, put it inside a method and use the method.

experiments/nf_evaluation/run_experiments.py
```diff
@@ -0,0 +1,67 @@
...
+from neuralforecast.losses.numpy import mae, mse, smape, mape
+
+from models import get_model
+from datasets import get_dataset
+
+# For compatibitlity with Mac with M chip
+os.environ["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"
+
+
+def main(args):
```
VascoSch92, line 16: i will call it `run_experiments`

experiments/nf_evaluation/run_experiments.py
```diff
@@ -0,0 +1,67 @@
...
+    model_smape = smape(y=forecasts_df['y'], y_hat=forecasts_df[f'{args.model}'])
+    model_mape = mape(y=forecasts_df['y'], y_hat=forecasts_df[f'{args.model}'])
+
+    metrics = {
+            'Model': [args.model],
+            'MAE': [model_mae],
+            'MSE': [model_mse],
+            'sMAPE': [model_smape],
+            'MAPE': [model_mape],
+            'time': [elapsed_time]
```
VascoSch92, line 44: 'elapsed_time' instead of time?

experiments/nf_evaluation/run_experiments.py
```diff
@@ -0,0 +1,67 @@
...
+    }
+        
+    
+    # Save results
+    results_path = f'./results/{args.dataset}'
+    os.makedirs(results_path, exist_ok=True)
+
+    metrics_df = pd.DataFrame(metrics)
```
VascoSch92, line 52: for every model you are creating a pandas dataframe. What if you loop evaluate every model and you create a pandas dataframe with all the results? It will be much easier to compare different models

VascoSch92 (commented):
Instead of passing the dataset name as an argument, you could use a config file? (yaml, toml or whatever).

In this way if you want to add another dataset you have just to change the config and not three scripts ;-)

Issue #852, opened 2024-01-08
How to Use  Different Optimizers with NeuralForecast Models


Amirh63:
### Description

Greetings Gentlemen,

I am currently working with the NeuralForecast library and specifically interested in training models using the Adam optimizer. I've gone through the documentation but haven't found clear guidance on how to integrate a custom optimizer, like Adam, into the training process.

Could you please provide some insights or examples on how to set up the Adam optimizer for training NeuralForecast models? I'm particularly interested in any necessary configurations or modifications needed within the library's framework to achieve this.

Additionally, if there are any best practices or recommendations for using custom optimizers with NeuralForecast, that information would be greatly appreciated.

Thank you for your time and assistance.

candalfigomoro:
Adam is the optimizer that is already being used

Amirh63:
How do I use a different optimizer other than Adam?

cchallu:
Hi @Amirh63. There is currently no option to modify the optimizer using the hyperparameters. We fixed the optimizer to Adam, given that all papers proposing these models use it. But this can be a great addition, so we will add it to our future improvements. In the meantime, you can modify it manually if you clone the repo and modify the following:

In line 181 on _base_windows.py:
```
def configure_optimizers(self):
        optimizer = torch.optim.Adam(self.parameters(), lr=self.learning_rate)
        scheduler = {
            "scheduler": torch.optim.lr_scheduler.StepLR(
                optimizer=optimizer, step_size=self.lr_decay_steps, gamma=0.5
            ),
            "frequency": 1,
            "interval": "step",
        }
        return {"optimizer": optimizer, "lr_scheduler": scheduler}
```

cchallu:
@quest-bot stash 200

cchallu:
Solution:
- Add `optimizer` hyperparameter on all models. The default should be None, and it recieves a custom `torch.optim` object.
- If user adds an `optimizer`, pass it to the `configure_optimizers` function (see above) and use instead of default. Repeat for all base classes.
- Add a test on `core.ipynb` that when user passes a different optimizer, results are different.

helenehanyu:
Hi @cchallu , can you make scheduler an option to?

cchallu:
@quest-bot reward @JQGoh

PR #901, opened 2024-02-25
Add option to support user defined optimizer for NeuralForecast Models

JQGoh:
This addresses https://github.com/Nixtla/neuralforecast/issues/852 regarding the use of a customized optimizer choice for Neuralforecast models.


neuralforecast/common/_base_multivariate.py
```diff
@@ -151,7 +160,12 @@ def on_fit_start(self):
         random.seed(self.random_seed)
 
     def configure_optimizers(self):
-        optimizer = torch.optim.Adam(self.parameters(), lr=self.learning_rate)
+        if self.optimizer:
+            optimizer = self.optimizer(
+                params=self.parameters(), **self.optimizer_kwargs
+            )
+        else:
+            optimizer = torch.optim.Adam(self.parameters(), lr=self.learning_rate)
```
AzulGarza, line 168: since `learning_rate` is a common argument across models, it might be better to catch if `self.optimizer` receives `lr` and update `self.optimizer_kwargs` to prevent confusion regarding the usage of this parameter in case a custom optimizer is used. for example, 

```python
optimizer_signature = inspect.signature(self.optimizer)
optimizer_kwargs = deepcopy(self.optimizer_kwargs)
if "lr" in optimizer_signature:
    if "lr" in optimizer_kwargs:
          warnings.warn("ignoring learning rate passed in optimizer_kwargs, using the model's learning rate")
     optimizer_kwargs["lr"] = self.learning_rate
```

nbs/core.ipynb
cchallu, line 1: Can we test it only on two or three models and max_steps=1, please?

cchallu: yes, sounds good!

cchallu (approved):

PR #907, opened 2024-02-29
fix base_multivariate window generation

jmoralez:
Removes the shuffling from the train dataloader so that each serie is always in the same position and makes sure the sampling considers all possible windows.


AzulGarza (approved):
awesome @jmoralez! lgtm, thank you🙌

PR #908, opened 2024-03-01
Add Time-LLM

marcopeix:
Add TimeLLM to neuralforecast.

User can load an LLM from HugginFace and use it for time series forecasting.


nbs/models.timellm.ipynb
```diff
@@ -0,0 +1,940 @@
+{
```
cchallu, line 1: <pre class="ql-syntax" spellcheck="false">Line #74.                     llm = None,
</pre><p> </p><p>If it is needed, remove default of None. Also, it is not clear from the docstring how to use it, what do users need to pass?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/908/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

marcopeix: Done! I added GPT-2 as the default LLM if the user does not specify another model.

marcopeix:
I added tests to the notebook
GPT-2 is the default LLM is the user does not specify another model
The usage example acts as a tutorial for using Time-LLM
Added `transformers` as a dependency

experiments/test_timellm/test_timellm.py
```diff
@@ -0,0 +1,55 @@
+import os
```
AzulGarza, line 1: this test is also part of `models.timellm.ipynb`, could we remove this file to keep the unit tests in the nbs?

neuralforecast/models/timellm.py
```diff
@@ -0,0 +1,497 @@
...
+        n_heads: int = 8,
+        enc_in: int = 7,
+        dec_in: int = 7,
+        llm=None,
+        llm_config=None,
+        llm_tokenizer=None,
+        llm_num_hidden_layers=32,
+        llm_output_attention: bool = True,
+        llm_output_hidden_states: bool = True,
+        prompt_prefix: str = None,
```
AzulGarza, line 293: adding `propmpt_prefix: Optional[str] = None` should fix the [linter](https://github.com/Nixtla/neuralforecast/actions/runs/8142491133/job/22252164542?pr=908) errors.

AzulGarza (changes requested):
thanks @marcopeix! 

could you also include `transformers` in the `dev_requirements` part of the [`setttings.ini`](https://github.com/Nixtla/neuralforecast/blob/10bb2e2f861d511123ec95646c655f2403da5f92/settings.ini#L19) file? 

also running `nbdev_install_git_hooks ` and `nbdev_clean` should solve the [`nb_sync` action](https://github.com/Nixtla/neuralforecast/actions/runs/8142491138/job/22252165334?pr=908).

neuralforecast/models/timellm.py
```diff
@@ -0,0 +1,497 @@
...
+from typing import Optional
+
+import torch
+import torch.nn as nn
+
+from ..common._base_windows import BaseWindows
+
+from ..losses.pytorch import MAE
+
+from transformers import GPT2Config, GPT2Model, GPT2Tokenizer
```
AzulGarza, line 18: since `transfomers` is not a core dependency, we can try to see if it is installed and then throw an error when it needs to be used for default arguments (`llm=None`), something like:

```python
IS_TRANSFORMERS_INSTALLED = True
try:
     from transformers import GPT2Config, GPT2Model, GPT2Tokenizer
except ImportError:
     IS_TRANSFORMERS_INSTALLED = False

File: neuralforecast/models/timellm.py [hunk 5]
```diff
@@ -0,0 +1,497 @@
...
+    `patch_len`: int=16, length of patch.<br>
+    `stride`: int=8, stride of patch.<br>
+    `d_ff`: int=128, dimension of fcn.<br>
+    `top_k`: int=5, top tokens to consider.<br>
+    `d_llm`: int=768, hidden dimension of LLM.<br>
+    `d_model`: int=32, dimension of model.<br>
+    `n_heads`: int=8, number of heads in attention layer.<br>
+    `enc_in`: int=7, encoder input size.<br>
+    `dec_in`: int=7, decoder input size.<br>
+    `llm` = None, LLM model to use. For example, you can use GPT-2 from https://huggingface.co/openai-community/gpt2"<br>
```
AzulGarza, line 236: could we document the default behavior when `llm=None`?

neuralforecast/models/timellm.py
```diff
@@ -0,0 +1,497 @@
...
+        self.llm_tokenizer = llm_tokenizer
+
+        if self.llm_config is None:
+            self.llm_confg = GPT2Config.from_pretrained("openai-community/gpt2")
+        if self.llm is None:
+            self.llm = GPT2Model.from_pretrained(
+                "openai-community/gpt2", config=self.llm_confg
+            )
+        if self.llm_tokenizer is None:
+            self.llm_tokenizer = GPT2Tokenizer.from_pretrained("openai-community/gpt2")
```
AzulGarza, line 374: i think it might be better if we use `GPT2` when `llm is None`, I'm thinking for example, what happens if `llm_tokenizer is not None` but for some reason `llm is None`? also, we should add an exception based on `IS_TRANSFORMERS_INSTALLED` in case that the default params are being used, something like:

```python
if self.llm is None:
      if not IS_TRANSFORMERS_INSTALLED:
             raise ImportError("please install `transformers` to use default llm")
      print("Using GPT2 model as default and ignoring `llm_config` and `llm_tokenizer`")
      self.llm = GPT2...
      self.llm_config = GPT2...
      self.llm_tokenizer = GPT2...
```

cchallu:
I don't think llm = None is a good idea. Either put the default model or nothing, like h or input size so users know they have to fill it.

marcopeix:
Updated PR according to Azul's comments:
- Better documentation of default behaviour
- Remove transformers from core dependencies and added to dev dependencies
- Flag to check if `transformers` is installed and use GPT-2 as default LLM
- Remove test file from experiments folder since the script exists in the notebook

AzulGarza (approved):
awesome @marcopeix! thank you 🎉

let's wait for the review of @jmoralez or @cchallu to merge it to main❤️

cchallu (approved):

PR #914, opened 2024-03-05
[FEAT] Added TSMixer model

elephaint:
This PR adds the [TSMixer](http://arxiv.org/abs/2303.06053) model into neuralforecast, and demonstrates its applicaiton in a Multivariate_with_TSMixer example notebook.

- Added TSMixer and AutoTSMixer models to NeuralForecast
- Included tests in core class
- Included performance evaluation in example notebook.


tsmixer_additional/_old/test_tf_vs_pt.py
AzulGarza, line 1: nice comparison! 
could you remove the `tsmixer_additional` dir? since it's validation code, we can store it in a separate github repository. (to prevent having protected code from other organizations)

environment-cuda.yml
AzulGarza, line 1: wdyt @jmoralez? should we have a different env file for cuda?

AzulGarza (changes requested):
nice @elephaint! i left a couple of comments ;)

environment-cuda.yml
jmoralez: What's the purpose of making the current one CPU-only? A faster install time?

elephaint: @jmoralez the current env file is cpu-only - it installs cpu by default (which I felt was ambiguous) - so this makes it explicit and is in line with the canonical way of installing PyTorch-cpu according to the PyTorch website.

tsmixer_additional/_old/test_tf_vs_pt.py
elephaint: Ah yes, of course

environment-cuda.yml
jmoralez: I think it will install cpu-only or with CUDA depending on the machine, you can see the files [here](https://anaconda.org/conda-forge/pytorch/files). I created my local env from that file and running `conda list pytorch` I get:
```
# Name                    Version                   Build  Channel
pytorch                   2.1.0           cuda118py311h3b8d1c1_300    conda-forge
```

AzulGarza: omg i was diving into some resources and it turns out conda can indeed set up pytorch with either cuda support or just cpu, based on what the system can handle. i didn't know! xD i also read some info saying for specific cuda drivers, you might need to add them in explicitly. @elephaint, could this be what happened in your case?

thinking about whether to include the cuda environment or keep the version in main, i see pluses and minuses both ways. having a cuda-specific env makes it super clear for anyone contributing which env to use. but, it does increase complexity in maintaining the repo, especially when a new dependency needs to be included. my lean is towards keeping things as they are in the main version and maybe just making a note in the [contributing file](https://github.com/Nixtla/neuralforecast/blob/main/CONTRIBUTING.md) that sometimes, you might need to install `pytorch-cuda` too. wdyt?

elephaint: @AzulGarza Makes sense; for me the reasons to include it separately where:

1. We don't make the use of CPU/GPU explicit anywhere in the repo, so it's now 'up to the user' to find out what device she/he ends up using. There seems to be no user control over it (?).
2. The current environment file - with the existing order of the channels - installs PyTorch from conda-forge which is [not maintained by PyTorch](https://discuss.pytorch.org/t/is-pytorch-cuda-going-to-be-registered-on-conda-forge/173968). To solve this issue, I switched the order of conda-forge and pytorch channels in the channel list in the environment file. Thus, PyTorch will then be installed from the 'official' PyTorch conda channels. The [conda-forge channel has no windows wheels](https://anaconda.org/conda-forge/pytorch/files), therefore in my case (Windows 11) a generic cpu version was installed when following our instructions even though I have a cuda device. The [official PyTorch channel](https://anaconda.org/pytorch/pytorch/files) does have pre-build versions for most common operating systems.

I suggest then to:
1. Remove the env-cuda file
2. Update the existing environment file: switching the channel order and removing `cpuonly`
3. Let me check whether PyTorch installs cuda correctly on Windows with the updated environment file (i.e. with the order of the channels switched).

Wdyt?

jmoralez: I think none of us use windows so we hadn't noticed that. Given the information you shared I agree with having two separate files (we could maybe rename the cpu only to environment-cpu).

About the hardware control I usually set the visible devices env variable when I want to run something on cpu (given that I have the cuda version installed), e.g. `CUDA_VISIBLE_DEVICES='' nbdev_test` or `CUDA_VISIBLE_DEVICES='' jupyter lab`. In case it helps.

AzulGarza: 🥳

CONTRIBUTING.md
```diff
@@ -35,12 +35,15 @@ Bug fixes and features are added through pull requests (PRs).
...
+Once you have `conda` go to the top level directory of the repository and run the following lines:
+````
 conda create -n neuralforecast python=3.10
 conda activate neuralforecast
-conda env update -f environment.yml
+```
+Then, run one of the following commands:
+```
+conda env update -f environment-cpu.yml  # choose this if you want to install the CPU-only version of neuralforecast
+conda env update -f environment-cuda.yml # choose this if you want to install the CUDA-enabled version of neuralforecast
```
AzulGarza, line 46: nice

AzulGarza (approved):
LGTM, thanks @elephaint! 🎉

candalfigomoro:
@elephaint 
- If we have several thousands of time series (making multivariate forecasting unfeasible), is it possible to use it in an "univariate" way (e.g. would setting `n_series=1` work)?
- Does it correctly support `futr_exog_list`, `hist_exog_list`, and `stat_exog_list` for each time series (like in this implementation https://github.com/ditschuk/pytorch-tsmixer?tab=readme-ov-file#tsmixerext)?

Thanks

elephaint:
> @elephaint
> 
> * If we have several thousands of time series (making multivariate forecasting unfeasible), is it possible to use it in an "univariate" way (e.g. would setting `n_series=1` work)?

I wouldn't do it - we use the `unique_id` to distinguish individual time series and it would require overwriting this / setting it to the same value for all series in order to function with n_series=1. However, that will introduce a lot of other issues. So, personally I would not do it and then revert to another univariate method.

> * Does it correctly support `futr_exog_list`, `hist_exog_list`, and `stat_exog_list` for each time series (like in this implementation https://github.com/ditschuk/pytorch-tsmixer?tab=readme-ov-file#tsmixerext)?
> 
> Thanks

No, we will soon be releasing a separate version that supports exogenous inputs ('TSMixerx'). Unfortunately the version supporting exogenous variables is quite different from the original TSMixer. We have a working version of TSMixerx on the AirPassengers Panel data with exogenous inputs, I'm still  making sure it works with the rest of our library/examples. Hopefully ready for release next week.

candalfigomoro:
@elephaint Thanks! So basically also TSMixerx will not be suitable for "univariate" + exog variables forecasting (like we do with NHITS, NBEATSx, etc.)?

elephaint:
Correct

PR #915, opened 2024-03-05
extract common methods to BaseModel

jmoralez:
Defines a `BaseModel` class from which `BaseMultivariate`, `BaseWindows` and `BaseRecurrent` can inherit to remove duplicated code and make it easier to add functionality to all models in a single place. This is just a refactor.


AzulGarza (approved):
awesome refactoring @jmoralez! now the code is cleaner and it will be easier to change the default behavior for all kinds of models if needed.🎉

PR #920, opened 2024-03-08
[FEAT] Added TSMixerx model

elephaint:
This PR adds the [TSMixerx](http://arxiv.org/abs/2303.06053) model into neuralforecast (in the paper referred to as `TSMixerExt`), which provides support for exogenous inputs, and demonstrates its applicaiton in a Multivariate_with_TSMixer example notebook.

- Added TSMixerx and AutoTSMixerx models to NeuralForecast
- Included tests in core class
- Included performance evaluation in example notebook.
- Updated some codeparts from TSMixer for consistency with TSMixerx


candalfigomoro:
If TSMixerx does support exog variables while TSMixer does not, should something like this be added to TSMixer?

```
        # TSMixer does not support historic exogenous variables
        if hist_exog_list is not None:
            raise Exception("TSMixer does not support historic exogenous variables.")

        # ... same for futr and static
```

As in models that do not support exog variables (NBEATS, PatchTST etc.)

elephaint:
> If TSMixerx does support exog variables while TSMixer does not, should something like this be added to TSMixer?
> 
> ```
>         # TSMixer does not support historic exogenous variables
>         if hist_exog_list is not None:
>             raise Exception("TSMixer does not support historic exogenous variables.")
> 
>         # ... same for futr and static
> ```
> 
> As in models that do not support exog variables (NBEATS, PatchTST etc.)

You're right - thanks for the suggestion. I'll make it part of the tsmixerx PR.

PR #923, opened 2024-03-11
Reduce device transfers

elephaint:
This PR reduces the amount of device transfers when creating tensors throughout the library.


cchallu (approved):

neuralforecast/losses/pytorch.py
```diff
@@ -1654,11 +1654,10 @@ def sample(self, distr_args, num_samples=None):
         sample_idxs = torch.multinomial(
             input=weights, num_samples=num_samples, replacement=True
```
jmoralez, line 1655: maybe we can initialize an empty tensor on the device and provide it as out here. Did you test if that improved the performance?

jmoralez: I just ran a quick test and it's indeed faster.
```python
import time

import torch

N = 10_000
num_samples = 100

# cpu
weights = torch.rand(N)
start = time.perf_counter()
sample_idxs = torch.multinomial(
    input=weights, num_samples=num_samples, replacement=True
).to('cuda')
print(f'{1000 * (time.perf_counter() - start):.0f}ms')
# 140ms

# cuda
weights = weights.to('cuda')
start = time.perf_counter()
sample_idxs = torch.empty(num_samples, dtype=torch.int64, device=weights.device)
torch.multinomial(
    input=weights, num_samples=num_samples, replacement=True, out=sample_idxs
)
print(f'{1000 * (time.perf_counter() - start):.0f}ms')
# 38ms
```

jmoralez: Nevermind. The multinomial's output is on the same device as the input, so we can actually remove the `sample_idxs = sample_idxs.to(probs.device)` line as well

PR #927, opened 2024-03-13
[FIX] 926 auto ensure all models support alias and 924 Configuring hyperparameter space for Auto* Models

### LINKED ISSUE #924: Configuring hyperparameter space for Auto* Models
### Description

As of now, users can pass a hyperparameter space using `config` argument which overrides the existing parameter grid. A new option `update_config` can be added so that users can make minor modifications in the default space.

### Use case

Quite frequently the default hyperparameter space is good enough, but sometimes users may want to control it in a minor way. For example, user may want to fix the context length to a specific value or may want to restrict the number of layers. As of now, passing a dictionary corresponding to only that field is behaving as if that's the entire search space along with default values of other parameters, but users intention is to restrict only few specific dimensions in the default space.

### LINKED ISSUE #926: [Auto] Ensure all models support alias

The BaseAuto class supports having an alias for the auto models, however not all of them expose it through their `__init__`, e.g. [AutoLSTM](https://github.com/sktime/sktime/pull/6107#discussion_r1520120033).

We should add a test to all auto models verifying that their signatures have all the expected arguments that the auto models have in common.

### Versions / Dependencies

1.6.4

### Reproduction script

```python
from neuralforecast.auto import AutoLSTM

AutoLSTM(h=1, alias='hi')
```

### Issue Severity

None

elephaint:
Fixes #926 and fixes #924.

- Fix of #926:
Fixed Auto* models that do not support all required variables from BaseAuto
Add unit test to all Auto* models to test if the Auto* models contain all required variables from BaseAuto
Some reformatting of models.ipynb, putting unit tests per model in hidden cell below each model
- Fix of #924:
Added a `get_default_config` method for all Auto* models
Fixed incompatible hyperparameter configurations in default_config and in three base models
Added unit tests to all auto models to ensure they all run with default configs, on both backends.
Update of Hyperparameter tuning guide to demonstrate the use of `get_default_config`


nbs/models.ipynb
```diff
@@ -83,6 +83,7 @@
     "\n",
```
jmoralez, line 83: <pre class="ql-syntax" spellcheck="false">Line #48.            reconciliation = config.pop('reconciliation')
</pre><p>It isn't possible to do this check with optuna, so this model doesn't suport it. Can you please add a check for the backend in the init and if it's not ray raise an error? Or if you can make it work that'd be awesome, I just couldn't think of a way to remove this key from the function body.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/927/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='97'/>

jmoralez, line 83: <pre class="ql-syntax" spellcheck="false">Line #47.                    config['n_series'] = n_series
</pre><p>I think this would work as well if you set it before the <code>if backend == 'optuna':</code> and then we can get rid of the <code>if backend == 'ray':</code>below. Same for the other ones</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/927/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='102'/>

elephaint: I think this boils down to essentially the same issue as your other remark: Optuna just doesn't work with Multivariate models and AutoHINT, as we need to enforce some parameters in the config. Let me think what would be best to solve it.

jmoralez: Ah I see, I thought we were only setting it for the default config. We should probably remove it from the auto model then so that the user always provides it through the config (and an error will be raised if it's not provided). WDYT?

elephaint: I included an exception for now, I haven't found a way to get Optuna to work, it fails in base_auto on `metrics = trainer.callback_metrics`, which is `None` when using the redefined `_fit_model` of `AutoHINT` (and haven't figured out how to correct this.)

elephaint: I came up with a method of checking the supplied config for `'n_series' = n_series` in case of Optuna backend, by using the MockTrial you created earlier. It raises an Exception if the user does not supply  `'n_series' = n_series`.

elephaint: Sorry, I missed your last comment (should have renewed the page). Yes, I can change it so that it always throws an Exception (now it only throws an Exception when using Optuna).

wdyt?

jmoralez: Removing it from the init signature is a breaking change, but I think that's the better option for the future. We can deprecate it and issue a FutureWarning indicating that the n_series should be provided through the config if it's provided in the init signature for the multivariate auto models.

jmoralez: Oh that should be easy to fix, that's the second argument of the return from _fit_model in AutoHINT, so you should go to HINT and store the trainer that is returned from the self.model.fit and propagate that back. I know it's hideous haha but I'm already working on improving it.

elephaint: @jmoralez In general, why do we ever want the user to supply `n_series` in MultiVariate models? We can infer it ourselves from `unique_id` in the dataset, right? In what situation is it not true that `n_series == unique(unique_id)`?

(This is more in the sense that if we make a breaking change to remove it, why not remove it entirely across all MultiVariate models)

elephaint: Fiddled around with it a bit but think it's indeed better to remove `n_series ` everywhere (i.e., also from the base multivariate models) rather than only from Auto* models, if we want to make that breaking change. 

wdyt?

elephaint: Ok, then I suggest we leave it as it is now (also for Auto*)

jmoralez: Agree

jmoralez (approved):
LGTM! Let's just please also wait for @cchallu's review as well.

elephaint:
> LGTM! Let's just please also wait for @cchallu's review as well.

Thanks - it does not fix the default_config issue yet, so I'm changing a few things (converted to draft)

elephaint:
- Added a `get_default_config` method for all Auto* models
- Fixed incompatible hyperparameter configurations in base models
- Added unit tests to all auto models to ensure they all run with default configs, on both backends.

nbs/examples/Automatic_Hyperparameter_Tuning.ipynb
```diff
@@ -181,11 +181,23 @@
    "cell_type": "markdown",
```
jmoralez, line 181: <pre class="ql-syntax" spellcheck="false">Line #4.        config = {**nhits_default_config(trial)}
</pre><p>The function already returns a dict, so there's no need to unpack it, you can assign it to the config variable directly or unpack it and return it in a single statement like:</p><pre class="ql-syntax" spellcheck="false">def config_nhits(trial):
    return {
        **nhits_default_config(trial),
        'random_seed': trial_suggest_int(...),
        ...
    }
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/927/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='37'/>

jmoralez (approved):
Just an optional comment

nbs/examples/Automatic_Hyperparameter_Tuning.ipynb
elephaint: Thanks - agree although that solution implicitly requires users to perform the operation in exactly that order, i.e.

```
return {
    **nhits_default_config(trial),
    'random_seed': trial_suggest_int(...),
    ...
}
```
works, but this:

```
return {
    'random_seed': trial_suggest_int(...),
    **nhits_default_config(trial),
    ...
}
```
will not correctly update the parameters. That's why I thought to make the order and update step explicit.

cchallu (approved):

PR #940, opened 2024-03-26
[FEAT] Sliding predictions for an arbitrary DF

howarth:
I want to be able to run something similar to predict_insample but with an arbitrary df instead of the training data. I didn't see anywhere else where I could run sliding window predictions. I moved the bulk of `predict_insample` to `_predict_sliding` and then I have `predict_insample` call `_predict_sliding`. Similarly, `predict_sliding(df, ..)` calls `_predict_sliding` 

This is not complete, I don't think it will work with exogenous variables, and it definitely will not work with static variables. I will figure that out. I wanted to first get feedback on if anything I'm doing here is incorrect. 

I think this is the type of predictions we will want to do for many of the medical projects at Auton


elephaint:
Thanks for submitting a PR! 

What is the use case for this?

Insample predictions on different datasets than the training set are a bit strange, imho... these predictions can't be used as valid forecasts, and they can't be used to improve the model.... so not sure what a good use case is that would justify the change?

marcopeix:
Reopening. I think we get the point of this, but we need to think about the best way of doing it.

PR #941, opened 2024-03-26
remove print from TimeLLM import

jmoralez:
If a user doesn't have the transformers library installed and imports any model, the `'The transformers library is required for Time-LLM to work'` message is printed, since `neuralforecast/models/__init__.py` imports `TimeLLM`. This removes that print since trying to initialize the model without transformers installed raises the following error: ```ImportError: Please install `transformers` to use the default LLM```, which should be enough.


AzulGarza (approved):
LGTM

PR #951, opened 2024-04-01
fix cross_validation when the id is the index

jmoralez:
If the user provided the id as the index to the cross_validation method a KeyError was raised because it'd always look for it in the columns. This adds the same check as for the other methods where a FutureWarning is issued but it still works.


AzulGarza (approved):
nice @jmoralez, lgtm ;)

PR #960, opened 2024-04-09
FEAT - Add iTransformer to multivariate forecasting example

marcopeix:
Added results from iTransformer (TSMixer still wins on ETTm2)
Added reference for iTransformer
Rename notebook to a more general title


nbs/examples/Multivariate_forecasting.ipynb
```diff
@@ -312,7 +312,7 @@
    "source": [
```
cchallu, line 312: <p>I think the iTransformer might be printing this, we should remove it from the model.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/960/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

marcopeix: <p>I actually don't know why this happens, but it's not related to iTransformer. For some reason, the kernel always crashes when I run this notebook, either on local or on a VM. However, Olivier can run it fine. I really don't know what could cause this.</p><p> </p>

marcopeix: In my case, the kernel dies because I run out of memory

elephaint: @marcopeix Likely related, but today I also had multiple kernel crashes when I ran iTransformer (as part of a tiny benchmark). Only happened with iTransformer. It's not a memory issue; it's a cuda kernel error (my screen blanked for a second too). It feels like the sort of error from asynchronously operating on a piece of data.

marcopeix: Ah, so weird, removing iTransformer does not solve the issue for me! Then, yes, we'll have to debug this, see why it happens!

marcopeix: After reviewing the issue, talking with Olivier and José, this seems to be an issue related to PyTorch and MacOS. I tried setting `accelerator='cpu'` and `device='cpu'` and it still did not run on Mac. We just have to run the notebook on another OS and it will remove the crashed kernel error message.

cchallu: Ok. I will pass the PR for the release tomorrow. We can run it again next week.

marcopeix:
Outdated PR, work was done in another PR.

Issue #66, opened 2021-06-28
EVALUATE: Weighted metrics have numerical error?


kdgutier:
During hyperopt evaluation breaks with:
np.average weights sum to zero can't be normalized

The error does not seem to come from a mask full of zeros.

mergenthaler:
@all-contributors please add @kdgutier  for code

PR #964, opened 2024-04-10
[FIX] HINT not producing coherent forecasts

### LINKED ISSUE #945: HINT outputs not Coherent

When running HINT, I expect the sum of lower level forecasts to be coherent with higher level forecasts. That is not happening.

### Versions / Dependencies

1.7.0

### Reproduction script

[Google Colab Demonstration of error](https://colab.research.google.com/drive/18CUa9eSyombCAa8Tl8HdJqg4AMchEIoe#scrollTo=kioJ9NJEeDkd)

### Issue Severity

elephaint:
In `HINT`, when we make predictions, we compute quantiles of the forecasts and subsequently calculate the mean using those quantiles. This gives an approximation error of the mean of the forecasts, leading a.o. to issue #945 but also showing the 'percentage difference' errors sometimes in our test runs with `HINT`.

The fix is simple: calculate the mean of the forecasts _before_ computing the quantiles.

Summary:
- Fixes #945 
- Unit test modified in `HINT.ipynb` that fails without the fix.
- Note that this also fixes generically the small eps in our tests that we observed when using `HINT`.


nbs/models.hint.ipynb
```diff
@@ -366,12 +366,15 @@
     "        \n",
```
jmoralez, line 366: <pre class="ql-syntax" spellcheck="false">Line #66.            assert np.allclose(children_sum, parent_value)
</pre><p>I think we should use <code>np.testing.assert_allclose</code>, since that has a lower tolerance and we're looking to get as close as possible here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/964/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

elephaint: Ok, changed it, although it might give issues at some point given we're using float32 most of the time.

cchallu (approved):

PR #971, opened 2024-04-15
[FEAT] TiDE model

elephaint:
Adds [TiDE model](http://arxiv.org/abs/2304.08424). 

Note:
- This implementation does not include a revin transformation explicitly (so users should do it via `scaler_type='revin'`).
- No categorical embeddings.


cchallu:
We are missing adding TiDE to the automatic evaluation in test-model-performance.

This should be added in the `action_files` folder, concretely: https://github.com/Nixtla/neuralforecast/blob/main/action_files/test_models/src/models.py and https://github.com/Nixtla/neuralforecast/blob/main/action_files/test_models/src/evaluation.py

elephaint:
> We are missing adding TiDE to the automatic evaluation in test-model-performance.
> 
> This should be added in the `action_files` folder, concretely: https://github.com/Nixtla/neuralforecast/blob/main/action_files/test_models/src/models.py and https://github.com/Nixtla/neuralforecast/blob/main/action_files/test_models/src/evaluation.py

@cchallu A lot of models are missing those scripts, and most models seem to be unused in models.py. Shouldn't all the models be included in the models list in models.py and evaluation.py? (because then I will make that change too)

cchallu:
Sorry, I forgot to mention. The idea is to have all univariate models in this evaluation, and we should include them at the moment of creating the model to assess their performance and detect potential bugs.

We should then have a separate script for multivariate models on a different dataset tailored for multivariate forecasting, like Ettm2.

B

elephaint:
> Sorry, I forgot to mention. The idea is to have all univariate models in this evaluation, and we should include them at the moment of creating the model to assess their performance and detect potential bugs.
> 
> We should then have a separate script for multivariate models on a different dataset tailored for multivariate forecasting, like Ettm2.
> 
> B

Ok, added BiTCN and TiDE to the script, so it should be ok for this PR then.

cchallu (approved):

PR #974, opened 2024-04-17
reduce default warnings

jmoralez:
Addresses the following warnings:
```python
pytorch_lightning/utilities/parsing.py:198: Attribute 'loss' is an instance of `nn.Module` and is already saved during checkpointing. It is recommended to ignore them using `self.save_hyperparameters(ignore=['loss'])`.
pytorch_lightning/trainer/connectors/data_connector.py:441: The 'val_dataloader' does not have many workers which may be a bottleneck. Consider increasing the value of the `num_workers` argument` to `num_workers=7` in the `DataLoader` to improve performance.
pytorch_lightning/trainer/connectors/data_connector.py:441: The 'train_dataloader' does not have many workers which may be a bottleneck. Consider increasing the value of the `num_workers` argument` to `num_workers=7` in the `DataLoader` to improve performance.
pytorch_lightning/trainer/connectors/data_connector.py:441: The 'predict_dataloader' does not have many workers which may be a bottleneck. Consider increasing the value of the `num_workers` argument` to `num_workers=7` in the `DataLoader` to improve performance.
pytorch_lightning/utilities/data.py:77: Trying to infer the `batch_size` from an ambiguous collection. The batch size we found is 1. To avoid any miscalculations, use `self.log(..., batch_size=batch_size)`.
neuralforecast/tsdataset.py:92: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
  self.temporal = torch.tensor(temporal, dtype=torch.float)
neuralforecast/tsdataset.py:96: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
  self.static = torch.tensor(static, dtype=torch.float)
```
by doing the following:
* Wrapping the `save_hyperparameters` call in a context manager that ignores warnings because we do want to save the `loss` passed to the constructor.
* Adding `pl.disable_possible_user_warnings()` in the `core` module to remove the warnings related to the dataloaders.
* Passing the corresponding `batch_size` to each `LightningModule.log` call. This is used to weigh the losses across steps in a batch (for the step methods) and across processes (for the epoch methods) in distributed.
* Using a copy constructor instead of `torch.tensor(some_tensor)` in the `tsdataset` module. I don't follow the suggestion to use `detach` because these arrays/tensors are always on the CPU, but I could add it.

Also cleans some notebooks.


AzulGarza (approved):

neuralforecast/core.py
```diff
@@ -59,6 +60,11 @@
 )
 
 # %% ../nbs/core.ipynb 5
+# this disables warnings about the number of workers in the dataloaders
+# which the user can't control
+pl.disable_possible_user_warnings()
```
fabianbergermann, line 65: I think this addition requires to bump the [required version](https://github.com/Nixtla/neuralforecast/blob/e0ecff827338af875acba76cf9e97e936ac56849/settings.ini#L18) of `pytorch-lightning` from 2.0.0 to 2.1.0, because the function `disable_possible_user_warnings` was only introduced in 2.1.0; see [the according PR](https://github.com/Lightning-AI/pytorch-lightning/commit/e02bb391af61eb02cf104ae032549b20243ec8f9) for context.
Many thanks for the great package 🙌

jmoralez: Thanks! I recently saw this as well and thought about changing that line to what the function does
https://github.com/Lightning-AI/pytorch-lightning/blob/2064887b12dd934a5f9a2bf45897f29e3bfc74d1/src/lightning/fabric/utilities/warnings.py#L37
since that warning type has been there longer. Would you like to contribute that?

fabianbergermann: ... happy to contribute this minor issue [in this PR](https://github.com/Nixtla/neuralforecast/pull/1081) . Note that this is my first contribution - I hope all is fine

PR #990, opened 2024-05-06
[FEAT] DeepNPTS model

elephaint:
Adds [DeepNPTS model](https://arxiv.org/pdf/2312.14657)


jmoralez:
This PR is very strange, it's showing diffs that shouldn't be there. I'll push an empty commit to see if it fixes that.

nbs/models.deepnpts.ipynb
```diff
@@ -0,0 +1,373 @@
+{
```
jmoralez, line 1: <p>Why is this not being run? We should check that it runs successfully to avoid breaking it in the future, if it takes too long just set max_steps=1</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/990/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

elephaint: Ah, yes. So I started this model with a copy from `DeepAR`, where this is also included, and forgot to remove it in the end. I'll remove it, this model is fast, so shouldn't give an issue in terms of time.

jmoralez (approved):

Issue #134, opened 2022-02-04
Missing import pandas forecast method


FedericoGarza:
**Is your feature request related to a problem? Please describe.**
The nhits' forecast method needs pandas to run properly. I think this is a silent bug because in the nbs , pandas is imported to test the method but it is not exported.

**Describe the solution you'd like**
Importing pandas from the beginning should solve the problem.

<img width="966" alt="image" src="https://user-images.githubusercontent.com/10517170/152598324-291bb078-f44e-42b1-81ee-ac05c24135d7.png">

PR #999, opened 2024-05-09
Feature/kan experiment

marcopeix:
Add KAN model
Benchmark of KAN on M3 and M4 dataset against MLP and NBEATS


experiments/kan_benchmark/README.md
AzulGarza, line 1: awesome @marcopeix! could we also add time and the nhits model? thank you

neuralforecast/models/kan.py
```diff
@@ -0,0 +1,364 @@
...
+
+# %% ../../nbs/models.kan.ipynb 6
+class KANLinear(torch.nn.Module):
+    def __init__(
+        self,
+        in_features,
+        out_features,
+        grid_size=5,
+        spline_order=3,
+        scale_noise=0.1,
```
AzulGarza, line 25: cool

experiments/kan_benchmark/environment.yml
```diff
@@ -0,0 +1,10 @@
+name: kan_benchmark
+channels:
+  - conda-forge
+dependencies:
+  - numpy<1.24
```
AzulGarza, line 5: do we know why we are using this restriction on `numpy`? is it related to the model?

AzulGarza (dismissed):
very cool @marcopeix, i left a couple of comments. also i think it might be pretty awesome to release the `Auto` version as well. perhaps in a different pr, once this one is merged.

experiments/kan_benchmark/environment.yml
marcopeix: I actually just copied this from the .yml file from the n-hits experiment. I can remove it!

elephaint (changes requested):
Cool!

- I think we should also add an `AutoKANLinear` model in `models.ipynb`, I think users will like the ability to have it auto-optimize when they are playing with the architecture. You can just copy-paste most of the code from a comparable MLP-based architecture and change the names.
- I think `KANLinear` should also be added to `models.py` and `evaluation.py` in action_files/test_models/src

experiments/kan_benchmark/README.md
marcopeix: Done!

experiments/kan_benchmark/environment.yml
marcopeix: Removed restriction on numpy.

nbs/core.ipynb
```diff
@@ -91,7 +91,7 @@
     "    Informer, Autoformer, FEDformer,\n",
```
cchallu, line 91: <p>autokan missing</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/999/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

nbs/models.kan.ipynb
```diff
@@ -0,0 +1,867 @@
+{
```
cchallu, line 1: <p>Why have only 1 hidden layer? the paper states more layers can be stacked.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/999/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

nbs/core.ipynb
marcopeix: Fixed!

nbs/models.kan.ipynb
marcopeix: Done! Users can specify an int or a list to hidden_size and specify the number of hidden layers.

If an integer is used, it will assign the same size to all hidden layers.

If a list is used, we check to ensure it has the same number of elements as the number of hidden layers specified.

nbs/models.kan.ipynb
```diff
@@ -0,0 +1,811 @@
...
```
cchallu, line 1: <pre class="ql-syntax" spellcheck="false">Line #134.                self.hidden_layers = [self.input_size] + [self.hidden_size for _ in range(self.n_hidden_layers)] + [self.h]
</pre><p>I suggest that we avoid explicit for loops here.</p><p> </p><p>The second term can be simplified to <code>self.n_hidden_layer*[self.hidden_size]</code> to improve readability. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/999/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

cchallu:
@marcopeix we also need to add KAN to the evaluation pipeline in `https://github.com/Nixtla/neuralforecast/tree/main/action_files`. You can check with Olivier how to add it,

elephaint (approved):
Looks good, @marcopeix!

marcopeix:
@cchallu I already added KAN to the evaluation pipeline in action_files. Olivier checked it and says it's good! Let me know if I am missing something!

marcopeix:
This PR is getting pretty big. Is it possible to review it and merge it soon? @cchallu @AzulGarza. Thanks! 😄

candalfigomoro:
@marcopeix Convolutional KAN could also be considered, see: https://github.com/IvanDrokin/torch-conv-kan

experiments/kan_benchmark/README.md
```diff
@@ -0,0 +1,72 @@
+# KAN for Forecasting - Benchmark on M3 and M4 datasets
+
+[Kolmogorov-Arnold Networks](https://arxiv.org/abs/2404.19756) (KANs) is an alternative to the multilayer perceptron (MLP). In this experiment, we assess the performance of KANs in forecasting time series.
+
+We use the M3 and M4 datasets, which represents more than 102 000 unique time series covering yearly, quarterly, monthly, weekly, daily and hourly frequencies.
+
+While KANs reduce the number of parameters by 38% to 92% compared to the MLP, it also rarely performs better than the MLP in time series forecasting tasks. In this benchmark, N-BEATS still performs best across the vast majority of datasets.
```
elephaint, line 7: Cool experiment! We could think about doing the experiment on a similar FLOP / parameter budget as MLP (this is just my science head thinking, no need to do it now, maybe to keep in mind for future benchmarks / work)

neuralforecast/auto.py
```diff
@@ -1105,7 +1107,74 @@ def get_default_config(cls, h, backend, n_series=None):
 
         return config
 
-# %% ../nbs/models.ipynb 75
+# %% ../nbs/models.ipynb 74
+class AutoKAN(BaseAuto):
+
+    default_config = {
```
elephaint, line 1113: Shouldn't this contain variations of the key parameters of the model, e.g. grid_size, spline_order?

neuralforecast/models/kan.py
```diff
@@ -0,0 +1,435 @@
...
+import math
+
+import torch
+import torch.nn.functional as F
+
+from ..losses.pytorch import MAE
+from ..common._base_windows import BaseWindows
+
+# %% ../../nbs/models.kan.ipynb 6
+class KANLinear(torch.nn.Module):
```
elephaint, line 18: Maybe move KANLinear to common modules if we use it in two models? (if we intend to keep KANHITS?)

elephaint (changes requested):
Awesome! Maybe also here we need to clear the outputs of the notebook of the models.

neuralforecast/auto.py
marcopeix: Good point! Fixed!

neuralforecast/models/kan.py
marcopeix: Not keeping KANHITS, so I'll leave it here. Maybe we'll move it when we work on TKAN or TKAT.

elephaint (approved):
Nice

Issue #560, opened 2023-05-04
[BUG] Mismatch between `loss` and `valid_loss` quantiles


kdgutier:
If the quantiles of the `loss` and `valid_loss` do not match, we get the following errors:
```
RuntimeError: The size of tensor a (n_quantiles_valid) must match the size of tensor b (n_quantiles) at non-singleton dimension 2
```

To reproduce we can simply make this instantiation with mismatching quantiles.
```
nhits = NHITS(h=horizon,
              input_size=24,
              loss=GMM(n_components=10, level=[80,90]),
              hist_exog_list=['month'],
              max_steps=2000,
              early_stop_patience_steps=10,
              val_check_steps=50,
              scaler_type='robust',
              learning_rate=1e-3,
              valid_loss=sCRPS(quantiles=quantiles))
```

It is convenient to add an Raise.Exception protection to the `BaseWindows` and the `BaseRecurrent` models to check and protect the matching quantiles.

PR #1007, opened 2024-05-14
[FEAT] Implicit Quantile Networks

elephaint:
Features:
- Adds Implicit Quantile Loss as a loss function to facilitate Implicite Quantile Networks for all architectures. This implementation is based on the [GluonTS implementation](https://github.com/awslabs/gluonts/blob/61133ef6e2d88177b32ace4afc6843ab9a7bc8cd/src/gluonts/torch/distributions/implicit_quantile_network.py), following [the original Gouttes IQN paper](https://arxiv.org/pdf/2107.03743).
- Cleanup moving parts to `base_model` out of the individual architectures, removing exceptions out of individual architectures.
- Adds a number of class attributes to models that indicate whether the model supports exogenous.
- Fixes some architecture errors, mostly in *former models.

Tests:
- Test for a range of architectures on AirPassengers

References:
- [IQN](https://arxiv.org/abs/1806.06923)
- [IQN for forecasting](https://arxiv.org/pdf/2107.03743)


nbs/core.ipynb
```diff
@@ -784,6 +784,7 @@
     "            last_dates = self.last_dates\n",
```
cchallu, line 784: <pre class="ql-syntax" spellcheck="false">Line #560.            cols = self._get_model_names()  # Hack for IQLoss, think about better way
</pre><p>Can you briefly explain the hack in a comment? I believe this line is needed again because the quantiles are modified in the loss of the model above, right?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1007/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

test/test_iqloss.py
cchallu, line 1: The test looks great!

Can we also add only one fast model trained with the new loss in the performance evaluation on `action_files`? With q=0.5 it will be comparable with the other methods.

nbs/core.ipynb
elephaint: Yes, you're right, it's needed because the quantile names are modified during the call to `predict` when one wishes to predict for a certain quantile, but the col names are then already 'established' during the earlier call to self._get_model_names(), resulting in the incorrect column name if one has specified a quantile different from the previous quantile. I couldn't think of an elegant way of solving this (as it seems not trivial to adjust cols afterwards either - then calling this function again migh be easiest.)

test/test_iqloss.py
elephaint: Added `NBEATS` with `IQLoss()` to the `action_files` (on my machine results in the best performance across these models with standard IQLoss()...)

cchallu (approved):

PR #1011, opened 2024-05-16
[FEAT] Add MOMENT

elephaint:
Feature:
- Adds [MOMENT ](https://arxiv.org/pdf/2402.03885) to NeuralForecast, following [this implementation](https://github.com/moment-timeseries-foundation-model/moment/tree/main). It is implemented (following the reference implementation) as a univariate model that does not support exogenous variables.

NB: tests show similar performance as reported, but the model is relatively slow (large model, many finetuning steps required).


nbs/models.moment.ipynb
```diff
@@ -0,0 +1,1518 @@
+{
```
AzulGarza, line 1: <pre class="ql-syntax" spellcheck="false">Line #111.            model = MOMENTPipeline.from_pretrained(
</pre><p>perhaps instead of using <code>model</code> we can use <code>moment</code> to prevent confusion with the <code>model</code> parameter defined above. maybe, it might be a good idea to rename that parameter to <code>mode_name</code>.     </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1011/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

AzulGarza, line 1: <pre class="ql-syntax" spellcheck="false">Line #112.                "AutonLab/MOMENT-1-large", 
</pre><p>i think here it's <code>model</code> instead of  <code>AutonLab/MOMENT-1-large</code>.  <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1011/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

AzulGarza (changes requested):
awesome @elephaint! i left a couple of comments. i was also speaking with @cchallu and we think it might be better to make it univariate in this iteration (mainly because for datasets with a lot of time series, the model explodes in GPU). this model is natively univariate, so in that regard there is no problem. wdyt?

nbs/models.moment.ipynb
elephaint: Ahh yes, you are very right

elephaint: Good catch

elephaint:
> awesome @elephaint! i left a couple of comments. i was also speaking with @cchallu and we think it might be better to make it univariate in this iteration (mainly because for datasets with a lot of time series, the model explodes in GPU). this model is natively univariate, so in that regard there is no problem. wdyt?

You're right! Changed to `BaseWindows`.

nbs/models.moment.ipynb
```diff
@@ -0,0 +1,1082 @@
...
```
marcopeix, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    class PositionalEmbedding(nn.Module):
</pre><p>Can we reuse the class in common_modules?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1011/discussion/#comment-1691400404'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

elephaint: It's not the same, the MOMENT version requires an additional input and selection mechanism in the forward method depending on the model name....

elephaint:
I think we're probably not going to merge this one.

PR #1014, opened 2024-05-20
Make MOMENT aware of insample mask, and freeze params

mononitogoswami:
This commit: (1) add parameters to freeze MOMENT weights. By default all parameters except that of the forecasting head are frozen, and (2) makes MOMENT aware of insample_mask, such that it does not attend to padding time steps


AzulGarza (approved):
awesome! thank you @mononitogoswami

Issue #21, opened 2021-05-02
Datasets: Use sklearn LabelEncoder to declare S matrix


kdgutier:
It seems convenient to use sklearn built-in functions to declare the static variables on all the datasets.
Time considerations and code convenience should be taken into account.

PR #1015, opened 2024-05-23
[FEAT] Add example of modifying the default configure_optimizers() behavior (use of ReduceLROnPlateau scheduler)

JQGoh:
# Rationale
 * Users have asked about whether we can modify the default learning rate scheduler behaviors. See https://github.com/Nixtla/neuralforecast/issues/852#issuecomment-1961923861 and https://nixtlacommunity.slack.com/archives/C031M8RLC66/p1713516602654109
 * This provides an example of how to achieve a full control of `configure_optimizers()` behavior and how we can use `ReduceLROnPlateau` scheduler during the optimization
 * ~~Instead of adding arguments to all the models, passing from NeuralForecast class, we introduce a `set_configure_optimizers` function for the BaseModel class such that individual model can overwrite the default **configure_optimizers()** behavior~~
 * ~~Note that this requires users to specify both `optimizer` and `scheduler` to have an effective optimization configuration, otherwise it shall fallback to the default option.~~

~~This shall deprecate the earlier work introduced in: https://github.com/Nixtla/neuralforecast/pull/998~~


jmoralez:
Hey @JQGoh, thanks a lot for trying this approach! Although I think it's fine to add more arguments for now, we can give this a shot when we work towards v2.0.

neuralforecast/common/_base_model.py
```diff
@@ -385,45 +363,65 @@ def on_fit_start(self):
...
+        scheduler = {
+            "scheduler": torch.optim.lr_scheduler.StepLR(
+                optimizer=optimizer, step_size=self.lr_decay_steps, gamma=0.5
+            ),
+            "frequency": 1,
+            "interval": "step",
+        }
+        return {"optimizer": optimizer, "lr_scheduler": scheduler}
+
+    def set_configure_optimizers(
```
jmoralez, line 381: I was thinking of taking an argument in the constructor that has a function with the same signature as `configure_optimizers` and modify the method to be something like:

```python
    if self.config_optimizers is not None:
        return self.config_optimizers(self)
```
to reduce the friction for users, since they may already know how to define that method and we just want to let them kind of override it. WDYT?

jmoralez: I think it needs self for the model parameters at least, not sure if other hyperparameters as well

jmoralez:
Thanks. Option 1 sounds good.

nbs/docs/tutorials/21_configure_optimizers.ipynb
```diff
@@ -0,0 +1,302 @@
+{
```
jmoralez, line 1: <pre class="ql-syntax" spellcheck="false">Line #21.        NHITS(h=horizon, input_size=input_size, max_steps=100),
</pre><p>can you give these models aliases and then use <code>utilsforecast.plotting.plot_series</code> on the next cell? That will produce a single plot with both models which will make it easier to compare.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1015/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

jmoralez (approved):
Thanks!

nbs/docs/tutorials/21_configure_optimizers.ipynb
jmoralez: Great!

PR #1028, opened 2024-06-10
fix num_tasks in spark integration

jmoralez:
Moves the logic of the distributed processing to separate methods to make it easier to follow the original methods (fit and predict).

During this refactor I found and fixed the error here: https://github.com/Nixtla/neuralforecast/blob/0e43ae0e7bc42294be95e1f4ebb497d7de1817d7/neuralforecast/common/_base_model.py#L333

After this I reran and updated the sample notebook, which was missing from the sidebar, so this adds it as well.


nbs/common.base_model.ipynb
elephaint, line 1: I think the change is fine, but isn't it easier to just up the Pytorch requirement for neuralforecast going forward?

nbs/core.ipynb
elephaint, line 1: I understand we only support Spark dataframes here - do we intend to extend it to other types of distributed dfs in the future or do we want to write separate functions for that?

elephaint (approved):

nbs/common.base_model.ipynb
jmoralez: It's not worth it dropping support for an entire minor version (2.0) for just one small change.

nbs/core.ipynb
jmoralez: Each framework has its own way of performing DDP, see for example [dask](https://github.com/saturncloud/dask-pytorch-ddp) and [ray](https://docs.ray.io/en/latest/train/getting-started-pytorch-lightning.html). So we may be able to reuse the dataset but we'll have to define different training functions for each.

PR #1030, opened 2024-06-12
[FEAT] - Add VCformer model

marcopeix:
Add the VCformer model, a multivariate forecasting model that uses a Variable Correlation attention mechanism and implements a Koopman Temporal Detector module.

[Paper](https://arxiv.org/abs/2405.11470)
[Code](https://github.com/CSyyn/VCformer)


neuralforecast/models/vcformer.py
```diff
@@ -0,0 +1,477 @@
...
+        # Element-wise multiplication and batched inverse FFT
+        res = q_fft * k_fft  # [B, D, D, T/2+1]
+        corr = torch.fft.irfft(res, dim=-1)
+
+        # Mean across the time dimension
+        corr = corr.mean(dim=-1)
+
+        return corr
+
+    def forward(self, queries, keys, values, attn_mask, tau=None, delta=None):
```
elephaint, line 225: This seems quite inefficient, don't know if this is the original code but I think we could refactor this ourselves, i.e. get rid of the for-loops and replace the fft cross-correlation operator by a normal conv1d.

neuralforecast/models/vcformer.py
```diff
@@ -0,0 +1,477 @@
...
+        k_fft = torch.fft.rfft(keys, dim=-1)
+        q_fft = q_fft.unsqueeze(1)  # [D,1,T/2+1]
+        k_fft = torch.conj(k_fft.unsqueeze(0))  # [1,D,T/2+1]
+        res = q_fft * k_fft  # [D,D,T/2+1]
+        corr = torch.fft.irfft(res, dim=-1)
+        corr = corr.mean(dim=-1)
+
+        return corr
+
+    def optimized_compute_cross_cor(self, queries, keys):
```
elephaint, line 207: Unused?

neuralforecast/models/vcformer.py
```diff
@@ -0,0 +1,477 @@
...
+        self.K = None
+        self.K_step = None
+
+    def forward(self, z):
+        B, N, input_len, hidden_dim = z.shape
+        pred_len = input_len
+        z = rearrange(z, "b n pn m -> (b n) pn m")
+        x, y = z[:, :-1], z[:, 1:]
+
+        self.K = torch.linalg.lstsq(x, y).solution
```
elephaint, line 52: We probably can't avoid this but this is also quite inefficient

neuralforecast/models/vcformer.py
marcopeix: Good catch! Removing it.

PR #1036, opened 2024-06-12
remove empty windows during predict

jmoralez:


jmoralez:
Seems like it, but that's more focused on the train stage. We can use that PR but having the threshold as an argument to the create windows method or similar so that we can always provide 0 during the predict step, because predicting windows full of zeros is just wasting compute.

The ideal scenario in my opinion would be not even creating those windows at all, but that requires a refactor on the dataset which impacts all models and thus may be harder.

Issue #409, opened 2023-01-17
Possible to fit a TFT with a custom large scale dataloader?


jaycunningham-8451:
The documentation for neuralforecast appears geared toward the use of pandas data frames.

I'm working on a time series forecasting problem for which data cannot be fit into memory on a single machine, and I've used petastorm in the past for this purpose. Is it possible to prepare a dataset manually as a dataloader and feed it to a neuralforecast model for training?

I'm happy to do any modification necessary, I just want to make sure that it's reasonable. From a cursory reading of the source it appears that the data frame assumption is fairly well embedded (but I could be incorrect).

kdgutier:
Hi @jaycunningham-8451,

If you are interested in large data settings we would need to modify the dataloader class:
- https://nixtla.github.io/neuralforecast/tsdataset.html
- https://github.com/Nixtla/neuralforecast/blob/main/neuralforecast/tsdataset.py#L16

In particular one would need to modify the `__getitem__(self, idx)` method to load from directories rather than parsing chunks of a pandas panel dataframe.

We would be really interested in collaborating for such modifications. Is your dataset public?

kdgutier:
Hey @jaycunningham-8451,

As you mentioned, our dataset reads the entire series and feeds them into `WindowsBase` or `RecurrentBase` kind of models. 

The idea that comes to my mind, is to mimic what people do in the computer vision literature is to read single photos individually, it might be possible to transform the numpy parsing operations from the `TimeSeriesDataset.__getitem__`. 

Here is some [documentation Pytorch large data generator](https://stanford.edu/~shervine/blog/pytorch-how-to-generate-data-parallel).

This option would require a new method to create specialized files indexed by the unique ids, that would be easily parsed back by the `__getitem__`.

Is this similar to what you think?

kdgutier:
I have not worked yet on large enough dataset that require me to go beyond the local disks.
Would it be possible to try first a version of the loader that works on single series assuming that we read from disk.
I picture scaling that code towards reading from the cloud could be very easy.

I think the most important part would be to have a parsing function that creates such dataset.

JoaquinDF-UniLU:
Hey @all is there any update on that regard?
I am facing the same issues. I need a custom large scale dataloader.

Best.

jasminerienecker:
Hi, we’re also facing the same issue: our dataset contains thousands of timeseries, each of which can fit into memory however which when combined into a single pandas dataframe is too large.
[...]

Issue #566, opened 2023-05-10
Issues with N-BEATSx and N-HITSx component analysis


sdmishra123:
Hello,

I was able to get the Level/Trend/Seasonal components for my model (NBEATS without exogs) using the link you shared, but I need help to get the components for my model (NBEATSx with exogs).
I keep getting the error - "RuntimeError: maximum size for tensor at dimension 2 is 280, but the size is 336",
when I run this line of code y_hat = nbeatsx.decompose(dataset=dataset). 

I used the below links to generate the components.

https://github.com/Nixtla/neuralforecast/blob/main/nbs/models.nhits.ipynb
https://github.com/Nixtla/neuralforecast/blob/main/nbs/models.nbeats.ipynb

In the below image, what does identity represent in N-BEATS? 
Why do I see seasonality in the trend component?

![N-BEATS component Analysis](https://github.com/Nixtla/neuralforecast/assets/61155429/304a1bc7-7200-4180-8798-74ec169cdaa1)

I have one more question, is there a way to get the feature importance for the exogenous variables used in the N-BEATSx and N-HITS model?

kdgutier:
Hey @sdmishra123,

Thanks for using for trying out the NBEATS decomposition.

**Regarding the `RuntimeError: maximum size for tensor at dimension 2 is 280, but the size is 336`**
- For the moment we are yet to include the exogenous partial predictions, as the current version of NBEATSx is not performing the projections on the exogenous basis, rather its taking the exogenous data as inputs of the MLPs.
- We will get to work on this feature soon.

**What does identity represent in NBEATS?**
- As you may know the interpretable NBEATS model is performing a decomposition, learning projection coefficients into polynomial and harmonic functions. $y_{[t+1:t+h]} = \alpha Poly +  \beta Sin/Cos$. In the case of and identity basis all the decoder is allowed to be freely determined that is $y_{[t+1:t+h]} = [\theta_{1},\theta_{2},...,\theta_{h}]$


**Why do I see seasonality in the trend component?**
- Your NBEATS model might not have been properly trained, you may want to look to the `AutoNBEATS` model, to automatically explored learning rate and training steps. Here is the [AutoNBEATS documentation](https://nixtla.github.io/neuralforecast/models.html#autonbeats), and here is an [AutoNBEATS usage example](https://nixtla.github.io/neuralforecast/examples/longhorizon_with_nhits.html).

[...]

PR #305, opened 2022-10-27
Static variables

cchallu:
* Added static exogenous variables to all models.
* Refactor TFT only to include continuous variables.
* Added static_df input to the core.NeuralForecast methods.
* Include usage examples in all models using core.NeuralForecast class.

PR #1049, opened 2024-07-01
Add support for datasets that can't fit in memory

jasminerienecker:
As described in this issue: https://github.com/Nixtla/neuralforecast/issues/409

We assume the dataset is split across multiple parquet files - each parquet file corresponds to a single timeseries which is represented as a pandas dataframe. This PR creates a new Dataset class where the __getitem__ method reads the parquet file corresponding to that index, and the from_data_directory() method replicates the from_df() method.

I have added a test to end of core.ipynb that checks the forecasts using this distributed dataset are the same as when the dataset is directly passed in as a pandas dataframe.


nbs/common.base_model.ipynb
```diff
@@ -46,7 +46,7 @@
     "\n",
```
jmoralez, line 46: <pre class="ql-syntax" spellcheck="false">Line #115.                
</pre><p>Please remove this extra line</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

nbs/core.ipynb
```diff
@@ -35,7 +35,7 @@
   {
```
jmoralez, line 35: <pre class="ql-syntax" spellcheck="false">Line #2.    from typing import Sequence
</pre><p>Please move this import above with the other typing imports</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 35: <pre class="ql-syntax" spellcheck="false">Line #204.            dataframe_or_files_list: Optional[Union[DataFrame, SparkDataFrame, Sequence[str]]] = None,
</pre><p>Please restore this to <code>df</code>. Even though it's not a very good name anymore this is a breaking change. We can think of deprecating it in favor of another name in the future.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

nbs/tsdataset.ipynb
```diff
@@ -54,7 +54,7 @@
     "#| export\n",
```
jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #4.    from typing import List, Sequence, Optional, Union
</pre><p>Please place Sequence after Optional, we want to add isort linting and this would fail</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #33.                df = pd.read_parquet(self.files_ds.files[idx])
</pre><p>I think we could have the convention of saving the files sorted by time, that way we can just read the temporal cols and assign them directly to data, e.g. <code>data = pd.read_parquet(self.files_ds.files[idx], columns=self.files_ds.temporal_cols).to_numpy()</code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #92.            for file in files:
</pre><p>Reading the whole files just to get the unique id and last time will be very slow. I think we can use the parquet metadata to extract this, something like:</p>

<pre class="ql-syntax" spellcheck="false">import pyarrow as pa

meta = pa.parquet.read_metadata(file)
rg = meta.row_group(0)
col2pos = {rg.column(i).path_in_schema: i for i in range(rg.num_columns)}
uid = rg.column(col2pos[target_col]).statistics.min
last_time = rg.column(col2pos[time_col]).statistics.max
for rg in range(1, meta.num_row_groups):
    last_time = max(last_time, rg.column(col2pos[time_col]).statistics.max)
</pre><p>We can also take the temporal cols as an argument and making sure they're there (by checking they're in the col2pos dict), that'd also allow using only a subset of features from the file.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez:
Thanks a lot for your contribution @jasminerienecker, I left some comments.

nbs/tsdataset.ipynb
jasminerienecker: yes makes sense - I've updated this so the temporal cols should now be directly input into the fit method (as shown in the example in core.ipynb) and we're just using the metadata to extract the relevant information

jmoralez: <p>Second thought on this. Since the id will just be repeated it'd be nice to put it in the filename instead, so that we don't repeat it for the length of the series. As a bonus, if we make the filename something like <code>unique_id=id_0</code> then it'd also work when loading from a hive partitioned dataset.</p><p> </p><p>So can you please change this to something like <code>uid = Path(file).name.split('=')[-1]</code>? </p>

nbs/tsdataset.ipynb
```diff
@@ -54,11 +54,12 @@
...
```
jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #81.            min_size = int('inf')
</pre><p>I think float('inf') as fine here, I get an error when trying to run this</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #82.            last_times = np.array([], dtype=np.datetime64)
</pre><p>Please make this a regular list. We also support having integer times, so I think it's better to use a regular list, append all the last times and then call np.array on that (should also be faster as well)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #83.            ids = np.array([])
</pre><p>Same for this, we should have a list, append to it and call pd.Index on it at the end.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #93.                for col in temporal_cols:
</pre><p>This could be done by making the temporal cols a set outside the loop <code>temporal_cols = set([target_col] + temporal_cols)</code> and then checking with <code>missing_cols = temporal_cols - col2pos.keys()</code>, which would allow us to display all the missing columns in the error message. Please also include the filename for easier debugging, e.g.</p>

<pre class="ql-syntax" spellcheck="false">if missing_cols:
    raise ValueError(f"Temporal columns: {missing_cols} not found in the file: {file}.")
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #97.                total_rows = rg.num_rows
</pre><p>If we enforce the convention of saving the series sorted by time we could also simplify this a bit by getting the maximum of the last row group:</p>

<pre class="ql-syntax" spellcheck="false">last_time = meta.row_group(meta.num_row_groups -1).column(col2pos[time_col]).statistics.max
</pre><p>Also we could get the total rows with this:</p>

<pre class="ql-syntax" spellcheck="false">total_rows = sum(meta.row_group(i).num_rows for i in range(meta.num_row_groups))
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez:
Thanks a lot for working through the changes! I left some more enhancement ideas

nbs/tsdataset.ipynb
jasminerienecker: Makes sense - I've even just passed that list directly into the pd.Index() structure, so avoided the explicit conversion to numpy

jasminerienecker: Sounds good, I've just kept it as a pd.Series though to make the type consistent with the indices in the from_df() method in the TimeSeriesDataset

jasminerienecker: I left the target_col out of the set to make sure it is the first entry in the temporal_cols variable (which seems to be an assumption later on)

nbs/tsdataset.ipynb
jasminerienecker: Thanks for all the suggestions! I missed this in the first commit but have added it as part of the most recent commit

nbs/common.base_model.ipynb
jmoralez: <p>That line is still there.</p><p> </p>

nbs/tsdataset.ipynb
jmoralez: <p>I'm sorry to do this again but I think the best format would be one that is compatible with hive partitioning, that'd make it very easy to take an existing dataframe and partition it with pandas for example <code>df.sort_values('ds').to_parquet('some_dir', partition_cols=['unique_id'])</code> or a very big dataset with spark:</p>

<pre class="ql-syntax" spellcheck="false">(
  df
  .repartitionByRange(numPartitions,"unique_id").
  .sortWithinPartitions("ds").
  .write.parquet("my_series.parquet")
)
</pre><p>So can you please change the logic to take a list of directories (which will be named: [unique_id=id_0, unique_id=id_1, ...]) where each one will hold one or more parquet files? <code>pd.read_parquet('unique_id=id_0')</code> will read all the partitions, so we just need to update the logic to get the metadata, I think something like this would work:</p>

<pre class="ql-syntax" spellcheck="false">for serie in series:
    serie_path = Path(serie)
    if not serie_path.is_directory():
        raise ValueError(f'paths must be directories, {serie} is not.')
&nbsp;&nbsp;&nbsp;&nbsp;uid = serie_path.name  # or serie_path.name.split('=')[-1]
    total_rows = 0
&nbsp;&nbsp;&nbsp;&nbsp;for partition in serie_path.iterdir():
        # add rows from partition and check it has the columns
</pre>

nbs/core.ipynb
jmoralez, line 35: <pre class="ql-syntax" spellcheck="false">Line #209.            temporal_cols: List[str] = [],
</pre><p>Sorry I didn't specify where to put this. We can infer what the temporal cols should be from the models, we already have this method in core <a href="https://github.com/Nixtla/neuralforecast/blob/0b9fad9d4bea56f4f278687fb03cb424ce25eb6f/neuralforecast/core.py#L542" target="_blank">https://github.com/Nixtla/neuralforecast/blob/0b9fad9d4bea56f4f278687fb03cb424ce25eb6f/neuralforecast/core.py#L542-L545</a> and we use a similar logic to get the historic exog. Then the temporal_cols will be the union of those</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 35: <pre class="ql-syntax" spellcheck="false">Line #222.            df : pandas, polars or spark DataFrame, or a list of files containing the DataFrame, optional (default=None)
</pre><p>Please change this to: <code>pandas, polars or spark DataFrame, or a list of parquet files containing the series, optional (default=None)</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 35: <pre class="ql-syntax" spellcheck="false">Line #305.                    f"<code>df</code> must be a list of parquet files, or a pandas, polars or spark DataFrame or <code>None</code>, got: {type(df)}"
</pre><p>Please move the list of files to the end to match the docstring (pandas, polars or spark DataFrame, or a list of parquet files containing the series)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 35: <pre class="ql-syntax" spellcheck="false">Line #566.            elif df is not None:
</pre><p>This can be just an if because the other conditions either return or raise an error</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 35: <p>Please remove this cell</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='62'/>

nbs/tsdataset.ipynb
```diff
@@ -54,13 +54,15 @@
...
```
jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #9.    import pyarrow as pa
</pre><p>Please move this import to the LocalFilesTimeSeriesDataset.from_data_directory method. We don't have a dependency on pyarrow and it'll only be needed when working with the local files dataset.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #12.    from pathlib import Path
</pre><p>Please move this import between collections and typing</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #32.            if isinstance(idx, int):
</pre><p>Can you please switch the order here to avoid indenting such a large portion? I realize that's the way the other method is currently defined but it's easier to read like:</p>

<pre class="ql-syntax" spellcheck="false">if not isinstance(idx, int):
    raise ValueError(f'idx must be int, got {type(idx)}')
# actual logic
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #34.                data = pd.read_parquet(self.files_ds.files[idx], columns=self.files_ds.temporal_cols.to_list()).to_numpy()
</pre><p>The columns argument can be an index, no need to call <code>to_list</code> on it </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #36.                # Add Available mask efficiently (without adding column to df)
</pre><p>I think this logic can be extracted to the base class</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #63.            # Define indexes if not given and then extract static features
</pre><p>I think this can be moved to the base class as well</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #102.                ids = ids + [uid]
</pre><p>Please use <code>ids.append(uid)</code> here, this copies the whole list every time</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #103.                last_times = last_times + [last_time]
</pre><p>append here as well</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #106.            ids = pd.Series(ids)
</pre><p>I think this needs to keep the name</p>

<pre class="ql-syntax" spellcheck="false">ids = pd.Series(ids, name=id_col)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <p>Please remove this cell</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

jmoralez:
Thanks a lot @jasminerienecker! I think after these changes it'll be ready

nbs/tsdataset.ipynb
jasminerienecker: No problem - I've updated this to support the dataframe being split between multiple files. I've had to re-add the logic to sort the dataframe though as we are now impacted by the order that the files are read in iterdir()!

nbs/core.ipynb
jasminerienecker: ah I see - I've now added this as the method _get_needed_temporal_cols in the _prepare_fit_for_local_files method

nbs/tsdataset.ipynb
jasminerienecker: I've now changed the logic here as we need to sort the dataframes in case the files within the directory aren't read in the right order

nbs/common.base_model.ipynb
jasminerienecker: ah yes that should be gone now!

nbs/tsdataset.ipynb
jmoralez: I'm not sure how to generate several files for a single partition in pandas, but spark names them as `part-00000`, `part-00001`, etc. which are read in that order by pandas when providing the directory only (we could also provide a list of the sorted filenames to be sure). So I think we can keep the `df = pd.read_parquet(self.files_ds.files[idx], columns=self.temporal_cols)` to avoid sorting on every read, which would make the loading very expensive.

Also, just in case more files are written, can you please change the iterator to `dir_path.glob('*.parquet')`

nbs/core.ipynb
jmoralez, line 35: <pre class="ql-syntax" spellcheck="false">Line #374.        def _get_needed_temporal_cols(self):
</pre><p>Please rename this to <code>_get_needed_exog</code>, since the temporal cols also include the target and the mask  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 35: <pre class="ql-syntax" spellcheck="false">Line #4.    df1_train = AirPassengersPanel_train[AirPassengersPanel_train['unique_id'] == 'Airline1']
</pre><p>Please use <code>AirPassengersPanel_train.to_parquet('data', partition_cols=['unique_id'], index=False)</code> instead of manually creating the directory to ensure compatibility. You can then define <code>data_directory = list(Path('data').iterdir())</code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='62'/>

nbs/tsdataset.ipynb
```diff
@@ -54,7 +54,8 @@
...
```
jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #44.        def _add_available_mask(data: np.ndarray, temporal_cols):
</pre><p>Please rename this to <code>_ensure_available_mask</code>, since it only adds it if it isn't already there  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #5.                     files_ds: _FilesDataset,
</pre><p>I don't think this needs to be a files dataset anymore, can be just a regular list of strings</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #35.            df = pd.read_parquet(self.files_ds.files[idx])
</pre><p>Please use</p>

<pre class="ql-syntax" spellcheck="false">data = pd.read_parquet(self.files_ds.files[idx], columns=self.temporal_cols).to_numpy()
</pre><p>here. Reading only the columns that we'll use in the order we'll use them should be very fast.</p><p> </p><p>For the available mask, can you please add that when reading the directories? You can add a flag or similar and if the first file has the mask then add it to the temporal cols set and check that it's present in every file. That way the final dataset will already read the mask.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #64.        def from_data_directories(directories, static_df=None, sort_df=False, temporal_cols=[], id_col='unique_id', time_col='ds', target_col='y'):
</pre><p>Please rename the temporal_cols argument here to <code>exogs</code> or similar to make it clearer that these are only the exogenous features</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 54: <pre class="ql-syntax" spellcheck="false">Line #77.            temporal_cols = set(temporal_cols)
</pre><p>This should also include the target</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1049/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jasminerienecker: I've now added available_mask to temporal_cols if it is already present in all the dataframes - I'm assuming this is what you meant? To be clear I still call         
        _data, temporal_cols = TimeSeriesDataset._ensure_available_mask(data, temporal_cols)_
in the get_items method as in _from_data_directories()_ we're only reading the metadata, not the full dataframe.

jasminerienecker: I've added the target at a later stage **temporal_cols = pd.Index([target_col] + list(exogs))** as later in the code there's the assumption that the target is the first index of temporal_cols. I can add it to the set too but I think I would then have to remove it and still call the above to make sure it is the first element in the list?

neuralforecast/tsdataset.py
```diff
@@ -370,12 +408,178 @@ def __init__(
...
+        # array with the last time for each timeseries
+        self.last_times = last_times
+        self.indices = indices
+        self.n_groups = len(files_ds)
+
+    def __getitem__(self, idx):
+        if not isinstance(idx, int):
+            raise ValueError(f"idx must be int, got {type(idx)}")
+
+        temporal_cols = self.temporal_cols
```
jmoralez, line 452: Please copy the index here, since that's later appended to
```suggestion
        temporal_cols = self.temporal_cols.copy()
```

neuralforecast/tsdataset.py
```diff
@@ -370,12 +408,178 @@ def __init__(
...
+        # Define indices if not given and then extract static features
+        static, static_cols = TimeSeriesDataset._extract_static_features(
+            static_df, sort_df, id_col
+        )
+
+        max_size = 0
+        min_size = float("inf")
+        last_times = []
+        ids = []
+        exogs = set(exogs)
```
jmoralez, line 502: You can use a different variable here just to check that the files have the expected temporal columns.
```suggestion
        expected_temporal = {target_col, *exogs}
```

neuralforecast/tsdataset.py
```diff
@@ -370,12 +408,178 @@ def __init__(
...
+                    max(last_time, last_time_file)
+                    if last_time is not None
+                    else last_time_file
+                )
+                total_rows += sum(
+                    meta.row_group(i).num_rows for i in range(meta.num_row_groups)
+                )
+
+                # Check all the temporal columns are present
+                missing_cols = exogs - col2pos.keys()
```
jmoralez, line 535:
```suggestion
                missing_cols = expected_temporal - col2pos.keys()
```

neuralforecast/tsdataset.py
```diff
@@ -370,12 +408,178 @@ def __init__(
...
+                    raise ValueError(
+                        f"Temporal columns: {missing_cols} not found in the file: {file}."
+                    )
+
+                if available_mask_seen and ("available_mask" not in col2pos.keys()):
+                    raise ValueError(
+                        f"available_mask column not found in the file: {file}, but found in previous files"
+                    )
+                elif "available_mask" in col2pos.keys():
+                    available_mask_seen = True
```
jmoralez, line 546: This will add it the first time it's seen and will check it in the following iterations.
```suggestion
                if "available_mask" in col2pos:
                    expected_temporal.add("available_mask")
```

neuralforecast/tsdataset.py
```diff
@@ -370,12 +408,178 @@ def __init__(
...
+            static_df, sort_df, id_col
+        )
+
+        max_size = 0
+        min_size = float("inf")
+        last_times = []
+        ids = []
+        exogs = set(exogs)
+        # Check if available_mask is present in all the directories
+        available_mask_seen = False
```
jmoralez, line 504:
```suggestion
```

neuralforecast/tsdataset.py
```diff
@@ -370,12 +408,178 @@ def __init__(
...
+            min_size = min(total_rows, min_size)
+            ids.append(uid)
+            last_times.append(last_time)
+
+        last_times = pd.Index(last_times, name=time_col)
+        ids = pd.Series(ids, name=id_col)
+
+        if available_mask_seen:
+            exogs.add("available_mask")
+        temporal_cols = pd.Index([target_col] + list(exogs))
```
jmoralez, line 558:
```suggestion
        if "available_mask" in expected_temporal:
            exogs = ["available_mask", *exogs]
        temporal_cols = pd.Index([target_col, *exogs])
```

jmoralez (changes requested):
Thanks! Just some minor improvements. The ReviewNB wasn't working for me so I hope you don't mind I put the comments in the code this time.

neuralforecast/tsdataset.py
jasminerienecker: Sounds good! I have also included the edge case that we see a number of files that don't contain the available_mask column before a file with the available_mask column present, as I don't think we want to add available_mask to temporal_cols in this case

```
if "available_mask" not in col2pos.keys():
        available_mask_seen = False
elif not available_mask_seen:
        # If this is triggered the available_mask column is present in this file but has been missing from previous files.
        raise ValueError("The available_mask column is present in some files but is missing in others." )
else:
        expected_temporal.add("available_mask")
```

jmoralez: Awesome!

neuralforecast/tsdataset.py
```diff
@@ -370,12 +408,180 @@ def __init__(
...
+        self.last_times = last_times
+        self.indices = indices
+        self.n_groups = len(files_ds)
+
+    def __getitem__(self, idx):
+        if not isinstance(idx, int):
+            raise ValueError(f"idx must be int, got {type(idx)}")
+
+        temporal_cols = self.temporal_cols.copy()
+        data = pd.read_parquet(self.files_ds[idx], columns=temporal_cols).to_numpy()
```
jmoralez, line 453: Sorry, it seems we do need this.
```suggestion
        data = pd.read_parquet(self.files_ds[idx], columns=temporal_cols.tolist()).to_numpy()
```

jmoralez (changes requested):

jmoralez:
Last request, it seems that the previous test in core is duplicated in the cell above, can you please remove that? Also please write to a temporary directory so that we don't leave files behind, e.g.
```python
with tempfile.TemporaryDirectory() as tmpdir:
    AirPassengersPanel_train.to_parquet(tmpdir, partition_cols=['unique_id'], index=False)
    data_directory = sorted([str(path) for path in Path(tmpdir).iterdir()])
    
    pred_df = AirPassengersPanel_train[AirPassengersPanel_train['unique_id'] == 'Airline2'].drop(columns='unique_id')
    futr_df = AirPassengersPanel_test[AirPassengersPanel_test['unique_id'] == 'Airline2'].drop(columns='unique_id')
    
    nf.fit(df=data_directory, use_init_models=True, id_col='id')
```
You can place the `import tempfile` in the cell with all the imports (the one that starts with)
```python
#| hide
import tempfile # <- place it here

import pytorch_lightning as pl

import neuralforecast
from ray import tune

from neuralforecast.auto import (
    AutoMLP, AutoNBEATS, 
    AutoRNN, AutoTCN, AutoDilatedRNN,
)

```

neuralforecast/tsdataset.py
jasminerienecker: Sounds good - when I was testing it seemed to work with the pd.Index but showed up as a typing error. This fixes that.

jasminerienecker:
Ah I completely missed that duplication. I've adjusted the test now and I think all the other comments have been resolved - let me know if you spot anything else though. 
In general I've been running this locally for model training and so far it seems to be working well!

jmoralez:
Can you check the failing test? There's a cell above that removes the trend column, you can try adding it again or using one of the features that the other cell introduces

jmoralez:
The CI is stuck, I'll try closing and reopening.

jasminerienecker:
@jmoralez ah right, I've moved the test to a more relevant section of the notebook as it seems by the end the train and test datasets didn't have aligned columns. It was working when I ran it locally so hopefully the tests will now pass in the PR too

jmoralez:
Can you try pushing an empty commit? Closing didn't work haha

jasminerienecker:
Hmm it still seems to be frozen...

jasminerienecker:
@jmoralez Sure here's a basic tutorial showing the expected format of the data, and briefly demonstrating fitting a model and predicting using this DataLoader https://github.com/Nixtla/neuralforecast/pull/1074 - happy to expand on some areas if it'd be useful but didn't want to duplicate too much with other tutorials!

shuyi-doordash:
Thanks a lot for providing this! I have a question, since I have >100K unique_ids, is there a way I can read them from parquets but not necessarily each unique_id on a single parquet? Besides, must that directory be a local one? I saved my data in s3 buckets..thanks!

PR #1055, opened 2024-07-03
[FEAT[ - Add multivariate evaluation to CI

marcopeix:
Add a step to evaluate multivariate forecasting models in the CI. Right now, it uses the ETTm2 dataset with a horizon of 96. Models tested are: TSMixer, SOFTS, iTransformer and AutoiTransformer. More models can be added.


elephaint (changes requested):
I think for the first iteration would be good to include all multivariate models, i.e. also StemGNN, MLPMultivariate and TSMixerx, and their Auto counterparts. Wdyt? 

I know the univariate testing is also incomplete, so not sure what the strategy is on that. I think we should test all models minimally with the CI on every change.

action_files/test_models/src/multivariate_evaluation.py
```diff
@@ -0,0 +1,65 @@
...
+    evals = pd.concat([evals, times], axis=1)
+
+    return evals
+
+
+if __name__ == '__main__':
+
+    groups = ['ETTm2']
+    
+    models = ['SOFTS',
```
elephaint, line 46: I'd add all multivariate models here

action_files/test_models/src/multivariate_models.py
```diff
@@ -0,0 +1,53 @@
...
+from src.data import get_data
+
+os.environ['NIXTLA_ID_AS_COL'] = '1'
+
+
+def main(dataset: str = 'multivariate', group: str = 'ETTm2') -> None:
+    train, horizon, freq = get_data('data/', dataset, group)
+    train['ds'] = pd.to_datetime(train['ds'])
+
+    models = [
```
elephaint, line 27: I'd add all multivariate models here

elephaint (changes requested):
Nice!

action_files/test_models/src/multivariate_evaluation.py
marcopeix: What model am I missing? I'm only including models inheriting BaseMultivariate.

action_files/test_models/src/multivariate_models.py
marcopeix: Same here, not sure which model I'm missing. We have SOFTS, TSMixer/TSMixerx, iTransformer, MLPmultivariate, StemGNN. Missing TimeMixer, but that PR is not merged yet.

elephaint (approved):

PR #1072, opened 2024-07-17
fixing bug in the NBEATSx exogenous basis stack

jasminerienecker:
Currently if you add an exogenous stack to the NBEATSx model when you have both future and static exogenous variables you get the error 

**RuntimeError: Tensors must have same number of dimensions: got 3 and 2**

This is because the code is trying to concatenate the future and static exogenous tensors which are of incompatible size.
I have added a test to the notebook that checks the model returns a prediction - if you remove my changes this test will fail.


elephaint (approved):
Awesome, thanks!

Issue #950, opened 2024-04-01
TimeLLM takes a long time to setup training.


hxuaj:
### What happened + What you expected to happen

Hi,
I was tring to run the example code of the TimeLLM model https://nixtlaverse.nixtla.io/neuralforecast/models.timellm.html#timellm
It took almost 1 hour before actual training. In the terminal, it only shows "Seed set to 1". I checked the GPU, where there is no GPU usage and only memory being taken about 500MB(~gpt2 size). Then the training began, it took only ~10s. In the training, there was usual GPU usage. Last, it also took ~1 hour to wrap up(predict time?).
I was wondering if it's a normal situation for TimeLLM since the model is new. If it's a problem, where the bottleneck could possibly be?
To exclude the network issue, I used local files to load GPT2:
```python
gpt2_config = GPT2Config.from_pretrained(gpt2_local_path, local_files_only=True)
gpt2 = GPT2Model.from_pretrained(gpt2_local_path, config=gpt2_config, local_files_only=True)
gpt2_tokenizer = GPT2Tokenizer.from_pretrained(gpt2_local_path, local_files_only=True)
```
Hardware: NVIDIA T4 (only tried it on one of my GPUs due to https://github.com/Nixtla/neuralforecast/issues/937)
OS: Linux

### Versions / Dependencies

Python 3.9
neuralforecast 1.7.0

### Reproduction script

https://nixtlaverse.nixtla.io/neuralforecast/models.timellm.html#timellm 

### Issue Severity

None

JKYtydt:
> ### 发生了什么 + 你期望发生什么
> 你好， 我试图运行 TimeLLM 模型的示例代码https://nixtlaverse.nixtla.io/neuralforecast/models.timellm.html#timellm 在实际训练之前花了将近 1 小时。在终端中，它只显示“Seed set to 1”。我检查了 GPU，没有 GPU 使用情况，仅占用了大约 500MB（~gpt2 大小）的内存。然后训练开始，只花了~10s。在训练中，GPU 使用情况很常见。最后，也花了大约 1 个小时来结束（预测时间？）。 我想知道这是否是 TimeLLM 的正常情况，因为该模型是新的。如果有问题，瓶颈可能在哪里？ 为了排除网络问题，我使用本地文件加载GPT2：
> 
> ```python
> gpt2_config = GPT2Config.from_pretrained(gpt2_local_path, local_files_only=True)
[...]

PR #1073, opened 2024-07-19
Fix issue #950: Reduce TimeLLM setup time for training

ive2go:
Regarding the issue (#950) with the TimeLLM model taking a long time, the root cause is as follows:

During the save_hyperparameters stage of the Pytorch Lightning module, the weights of the llm are saved, resulting in a long processing time and preventing the model from being moved to the GPU for training.

To address this issue, I have defined the llm model name in the TimeLLM definition and modified the TimeLLM model to load the llm weights internally. This prevents the weights from being saved during the save_hyperparameters stage.

The changes are limited to the timellm.py file.


jmoralez:
Hey @ive2go, thank you for your contribution! Can you please restore the changes on the examples?

ive2go:
@jmoralez I have restored the examples folder to its previous state. Please review the changes. Thank you.

jmoralez:
Thanks! Can you please also revert the changes to `nbs/models`, `nbs/models.softs` and the permissions change to `neuralforecast/_modidx.py`?

nbs/models.timellm.ipynb
```diff
@@ -68,7 +68,7 @@
     "from neuralforecast.losses.pytorch import MAE\n",
```
jmoralez, line 68: <pre class="ql-syntax" spellcheck="false">Line #157.                try:
</pre><p>This pattern is very awkward. Will it not look for local files first and only download if necessary?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1073/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

ive2go: Thank you for your comment. However, I believe there might be a misunderstanding here. This code is not actually downloading or loading any models. It's simply trying to import classes from the transformers library and checking if the library is installed.
The pattern used here is a common way to check for the availability of an optional dependency. It doesn't involve looking for local files or downloading anything. The actual model loading and potential downloading would happen later when these classes are used to instantiate models.
If you have concerns about how models are loaded or downloaded, that would be more relevant to the code where we actually use these classes to create model instances. Would you like to discuss that part of the code instead?

ive2go:
> Thanks! Can you please also revert the changes to `nbs/models`, `nbs/models.softs` and the permissions change to `neuralforecast/_modidx.py`?

Thank you for the feedback! I have reverted the changes to nbs/models.ipynb, nbs/models.softs.ipynb, and the permissions change to neuralforecast/_modidx.py as requested.

nbs/models.timellm.ipynb
jmoralez: The comment was made on the reviewnb app, not where github is showing it. I meant all the blocks where the only thing that changes is the `local_files_only` argument

neuralforecast/models/timellm.py
```diff
@@ -15,7 +15,17 @@
 from ..losses.pytorch import MAE
 
 try:
-    from transformers import GPT2Config, GPT2Model, GPT2Tokenizer
+    from transformers import (
```
elephaint, line 18: This fails if any of the imports fail. Why not only load GPT2?

neuralforecast/models/timellm.py
```diff
@@ -256,14 +266,12 @@ class TimeLLM(BaseWindows):
...
-    `d_llm`: int=768, hidden dimension of LLM.<br>
+    `d_llm`: int=768, hidden dimension of LLM.<br> # LLama7b:4096; GPT2-small:768; BERT-base:768
     `d_model`: int=32, dimension of model.<br>
     `n_heads`: int=8, number of heads in attention layer.<br>
     `enc_in`: int=7, encoder input size.<br>
     `dec_in`: int=7, decoder input size.<br>
-    `llm` = None, LLM model to use. If not specified, it will use GPT-2 from https://huggingface.co/openai-community/gpt2"<br>
-    `llm_config` = None, configuration of LLM. If not specified, it will use the configuration of GPT-2 from https://huggingface.co/openai-community/gpt2"<br>
-    `llm_tokenizer` = None, tokenizer of LLM. If not specified, it will use the GPT-2 tokenizer from https://huggingface.co/openai-community/gpt2"<br>
+    `llm_model` = GPT2, LLM model to use. If not specified, it will use GPT-2 from https://huggingface.co/openai-community/gpt2"<br> # LLAMA, GPT2, BERT
```
elephaint, line 274: Isn't it possible to re-use the existing variable names to prevent a breaking change? I'm fine with deprecating the llm_config and llm_tokenizer argument, but then we keep them and show users a deprecation message when they provide a non-default value

neuralforecast/models/timellm.py
```diff
@@ -319,9 +327,7 @@ def __init__(
         n_heads: int = 8,
         enc_in: int = 7,
         dec_in: int = 7,
-        llm=None,
-        llm_config=None,
-        llm_tokenizer=None,
+        llm_model: str = "GPT2",
```
elephaint, line 330: Same, ideally we don't deprecate existing arguments

neuralforecast/models/timellm.py
```diff
@@ -395,29 +401,122 @@ def __init__(
...
+        self.llm_model = llm_model
+        self.llm_num_hidden_layers = llm_num_hidden_layers
+        self.llm_output_attention = llm_output_attention
+        self.llm_output_hidden_states = llm_output_hidden_states
 
-        if self.llm is None:
-            if not IS_TRANSFORMERS_INSTALLED:
-                raise ImportError(
-                    "Please install `transformers` to use the default LLM"
+        if self.llm_model == "LLAMA":
```
elephaint, line 409: Essentially this limits users to three possible LLMs - isn't there an option where we keep the try/except construct but are able to feed it any LLM name?

elephaint (changes requested):
Thanks for all the work! See my comments

neuralforecast/models/timellm.py
jmoralez: I think saving those variables is what takes a long time and what this PR is for, I'd just suggest keeping the `llm` argument instead of also renaming it to reduce the scope of the breaking change.

ive2go:
### Note
We have proposed an alternative approach to address the same issue (#950) in a new PR(#1086). Both approaches aim to optimize the initialization time for the TimeLLM model, but they use different methods.

- **This PR**: 
  - Modifies the `timellm.py` file to handle the issue.
  - Loads the LLM weights internally within `TimeLLM` by passing only the model name and saves information about which LLM was selected in the hyperparameters.
  - Limitation: The approach restricts LLM options to three models (GPT-2, LLAMA, BERT).

- **Alternative PR(#1086)**:
  - Modifies the `_base_model.py` file to exclude LLM weights during the `save_hyperparameters` stage.
  - The LLM model information is not saved in the PyTorch logs, but it can be inferred from the config file or tokenizer.
  - Advantage: This approach provides flexibility in choosing any LLM model.

I personally prefer the new approach as it seems more flexible. However, I proposed the method in this PR because modifying the `_base_model.py` impacts not only the TimeLLM model but also other models that inherit from this parent class.

Reviewers can choose between these two solutions based on their preferences and the project's requirements. We appreciate your feedback and will respond to any additional suggestions or comments.

Thank you for reviewing both PRs.

ive2go:
> Thanks for all the work! See my comments

Thank you for your feedback. I have made the suggested changes throughout the code. Please review and let me know your thoughts.

nbs/models.timellm.ipynb
ive2go: I've made the changes based on your feedback.

neuralforecast/models/timellm.py
ive2go: I’ve updated it to load the default model(gpt2) if it fails to fetch.

neuralforecast/models/timellm.py
ive2go: I’ve updated it to reuse the existing variable names as you suggested.

neuralforecast/models/timellm.py
ive2go: I’ve updated it to reuse the existing variable names as you suggested.

neuralforecast/models/timellm.py
ive2go: I’ve updated it to use AutoModel, AutoTokenizer, and AutoConfig, so you can now use pretrained model paths beyond the three specific models.

marcopeix:
Hello @ive2go, can you solve the conflict in the notebook? I really want to move forward with this PR! Thanks!

ive2go:
> Hello @ive2go, can you solve the conflict in the notebook? I really want to move forward with this PR! Thanks!

Hi @marcopeix, I have removed the conflict markers and committed the changes, but the conflict still appears unresolved. Could you please check if I missed anything in the conflict resolution process? I want to ensure everything is correctly resolved so we can proceed with this PR. Thanks!

marcopeix:
For some reason, I was not able to update this PR, so I opened another PR with the sames changes and solved the conflicts there #1139

marcopeix:
Merged PR #1139 which fixes the conflicts in this PR. Thank you @ive2go for your work!

PR #1083, opened 2024-07-25
[Refactor] Move RevIN class to common module

JQGoh:
Rationale
 * To reduce code duplication
 * Also reuse `SeriesDecomp` class from common in fedformer, since they are the same.


neuralforecast/common/_modules.py
```diff
@@ -486,7 +487,87 @@ def forward(self, x, x_mark=None):
 
         return self.dropout(x)
 
-# %% ../../nbs/common.modules.ipynb 19
+# %% ../../nbs/common.modules.ipynb 20
+class Normalize(nn.Module):
```
elephaint, line 491: Can we name this class RevIN (as that's what it does)? Note that normalization isn't preferred in this location; ideally we have the normalization steps in common.scalers.ipynb (which also has revin), however for the case of RevIN specifically this has been somewhat hard to implement.

elephaint (changes requested):
Thanks! See my comment. I think it's good to collect common stuff in common.modules. 

Note that there are more models that use RevIN - please try to include all of them (from the top of my head at least also TSMixer and TSMixerx, but maybe more).

elephaint:
> > Thanks! See my comment. I think it's good to collect common stuff in common.modules.
> > Note that there are more models that use RevIN - please try to include all of them (from the top of my head at least also TSMixer and TSMixerx, but maybe more).
> 
> @elephaint Thanks for reminding me about the RevIN used in TSMixer and TSMixerx (think these are the ones I missed). However, I noticed that the implementation in those two classes said the weight initialization are
> 
> ```
> # TSMixer
> nn.Parameter(torch.ones((1, 1, n_series)))
> ```
> 
> and
> 
> ```
> # TSMixerx
> n.Parameter(torch.ones((1, 1, 1, n_series)))
> ```
> 
> and these do not share exactly the same implementation of `self.affine_weight` as suggested in this PR.
> 
> So I am thinking it could be more straight forward to keep the RevIN implementations for TSMixer and TSMixerx as it is. Let me know what do you think.

YOu're right, better to keep these as is

neuralforecast/common/_scalers.py
```diff
@@ -467,3 +467,83 @@ def inverse_transform(self, z, x_shift=None, x_scale=None):
     def forward(self, x):
         # The gradients are optained from BaseWindows/BaseRecurrent forwards.
         pass
+
+# %% ../../nbs/common.scalers.ipynb 38
+class RevIN(nn.Module):
```
elephaint, line 472: Given that we don't integrate RevIN as an actual scaler but keep it as a standalone module, I think it's best placed in `_modules` rather than `_scalers`

neuralforecast/common/_modules.py
elephaint: I see that I was a bit ambiguous here. In `_scalers`, we include scalers that scale the data prior to entering the NN, such as `standard` and `robust`. One could see `RevIN` as such a scaler (which I think it ultimately is). The comment above implied that if we collect all `RevIN` implementations, we could remove them from their original models and just have a user choose a scaler upon training the model, for which one of the options is `RevIN`. 

However, the current PR justs collects the existing RevIN modules into a common file, for which `_modules.py` is more appropriate. Makes sense what I mean? At the time of the comment I thought the idea of this PR was to fully integrate `RevIN` as a scaler (not as a module).

nbs/common.scalers.ipynb
```diff
@@ -777,6 +777,94 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "31549645",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| export\n",
+    "\n",
+    "class RevIN(nn.Module):\n",
```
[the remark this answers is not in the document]
marcopeix: I agree about placing it in `_modules` rather than in `_scalers`. Makes more sense the way it's implemented.

marcopeix:
Would love to get this merged before the end of the week and make a new release of neuralforecast.

marcopeix (approved):
LGTM! Thanks for the contribution!

elephaint (approved):
THanks!

PR #1088, opened 2024-07-29
FEAT: Support exog features in TimeMixer

marcopeix:
Fix bug that allows TimeMixer to take futr_exog now.


nbs/models.timemixer.ipynb
```diff
@@ -736,8 +729,8 @@
     "        futr_exog = windows_batch['futr_exog']\n",
     "\n",
     "        if self.futr_exog_size > 0:\n",
-    "            x_mark_enc = futr_exog[:,:self.input_size,:]\n",
-    "            x_mark_dec = futr_exog[:,-(self.label_len+self.h):,:]\n",
+    "            x_mark_enc = futr_exog[:, :, :self.input_size]\n",
+    "            x_mark_dec = futr_exog[:, -(self.h):, :]\n",
```
elephaint, line 733: This is incorrect, the shape of futr_exog is [B, F, L + h, N], so the selection in the sequence dimension should be in the third dimension.

marcopeix: Fixed!

elephaint:
@JQGoh 1.7.4 was already released today, we took out the futr_exogenous functionality of `TimeMixer` for now.

elephaint:
> > @JQGoh 1.7.4 was [already released today,](https://github.com/Nixtla/neuralforecast/releases/tag/v1.7.4) we took out the futr_exogenous functionality of `TimeMixer` for now, but we can make it part of the next release.
> 
> Thanks for the info. I was not aware that the `futr_exogenous` support was removed.

Yeah @marcopeix and I couldn't make it work easily and we didn't understand it sufficiently, and we wanted to move forward with the release...

peacocktrain:
@marcopeix  Hello, I would like to use the futr_exog feature of TimeMixer, and I have pulled down this branch to test it. However, the content provided in the ipynb file does not execute correctly, and the following error is reported:

`File "/workspace/neuralforecast/common/_modules.py", line 388, in forward
x = self.tokenConv(x.permute(0, 2, 1)).transpose(1, 2)
RuntimeError: permute(sparse_coo): number of dimensions in the tensor input does not match the length of the desired ordering of dimensions i.e. input.dim() = 4 is not equal to len(dims) = 3`

Could you please help me check if the modifications in the ipynb file can be executed correctly? Thank you.

elephaint:
@peacocktrain This is a draft PR, which means it doesn't necessarily work. Please don't use draft PRs to work with, as functionality likely doesn't work properly.

For now, I'd advise to use another model that supports futr_exog.

marcopeix:
Supporting exogenous features in TimeMixer is a feature that is on our radar. Note that the model as presented in the [original paper](https://openreview.net/pdf?id=7oLshfEIC2) does not use exogenous features. So, in the current state, the model in neuralforecast can be used to reproduce the results of the paper.

elephaint:
The original paper doesn't seem to support exogenous features. Upon learning this I don't see a reason to support this.

Note that it's not easy to include exogenous - it is insanely hard to do this right, so it seems a limitation of the architecture that we are not placed to solve.

I'd vote for closing unless I'm misunderstanding here.

PR #1095, opened 2024-07-30
Add support for the local file dataloader with the Automodels 

jasminerienecker:
Currently when we have exogenous variables and are using the LocalFilesDataloader, we cannot use the Automodels. This is because the _get_needed_exog() method doesn't pick up the exog variables defined in the config and so it throws an error.


neuralforecast/_modidx.py
```diff
@@ -500,60 +500,7 @@
                                                                                                          'neuralforecast/losses/pytorch.py'),
                                                'neuralforecast.losses.pytorch.weighted_average': ( 'losses.pytorch.html#weighted_average',
                                                                                                    'neuralforecast/losses/pytorch.py')},
-            'neuralforecast.models.autoformer': { 'neuralforecast.models.autoformer.AutoCorrelation': ( 'models.autoformer.html#autocorrelation',
```
jmoralez, line 503: can you please install "nbdev==2.3.25"? the latest version removes these symbols

neuralforecast/core.py
```diff
@@ -597,16 +597,23 @@ def get_missing_future(
         return ufp.anti_join(expected, futr_df[ids], on=ids)
 
     def _get_needed_futr_exog(self):
-        return set(
-            chain.from_iterable(getattr(m, "futr_exog_list", []) for m in self.models)
-        )
+        futr_exogs = []
+        for m in self.models:
+            futr_exogs += getattr(m, "futr_exog_list", [])
+            if len(getattr(m, "config", {})):
```
jmoralez, line 603: This probably won't work for optuna, I'll think on how to do this

jmoralez (changes requested):

neuralforecast/core.py
jasminerienecker: ah right yes I was just testing with ray - I'm not very familiar with Optuna so happy to modify according to suggestions

jmoralez: Hey. Sorry for taking so long. I think we can modify the MockTrial to return the choices
https://github.com/Nixtla/neuralforecast/blob/957a0ccb34d17395ad1da15ef1b027f8dc2c1944/neuralforecast/common/_base_auto.py#L23-L24
i.e. change it to:
```python
def suggest_categorical(self, name, choices):
    return choices
```
Then when trying to get the exogs we check if the config is a dict (ray) of if it's a callable (optuna)
* If it's ray we check if the features are being tuned, if they're just a list we extract them directly, otherwise the sample has a `categories` attribute that holds the possible values.
* If it's optuna we use the `MockTrial` to get the possible values and then extract them directly.

I believe something like this should work, but feel free to reach out if you get stuck:
```python
from neuralforecast.common._base_auto import BaseAuto, MockTrial

for m in self.models:
    if isinstance(m, BaseAuto):
        if isinstance(m.config, dict):  # ray
            exogs = m.config.get('futr_exog_list', [])
            if hasattr(exogs, 'categories'):  # features are being tuned, get possible values
                exogs = exogs.categories
        else:   # optuna
            exogs = m.config(MockTrial()).get('futr_exog_list', [])
    else:  # regular model, extract them directly
```

jasminerienecker: I've updated those scripts and that seems to work ok - I've also updated the tests to optuna to confirm this !

nbs/core.ipynb
```diff
@@ -94,7 +94,8 @@
     "    MLPMultivariate, iTransformer,\n",
```
jmoralez, line 94: <pre class="ql-syntax" spellcheck="false">Line #8.        AutoMLP(h=12, config=config_f, num_samples=2, backend='optuna')
</pre><p>Please also set the sampler so that the trials are the same across runs, i.e.</p>

<pre class="ql-syntax" spellcheck="false">AutoMLP(h=12, config=config_f, num_samples=2, backend='optuna', search_alg=optuna.samplers.TPESampler(seed=0))
</pre><p>Also please add another model with ray so that we test both backends and include exogenous in the tuning parameters, one constant and one being tuned, e.g.</p>

<pre class="ql-syntax" spellcheck="false">config = {
    'hist_exog_list': ['a'],
    'futr_exog_list': tune.choice([['b'], ['c'], ['b', 'c']])
}
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1095/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

jmoralez, line 94: <pre class="ql-syntax" spellcheck="false">Line #26.    pred_airline2 = pred_dataframe[pred_dataframe['id'] == 'Airline2'].reset_index(drop=True)
</pre><p>I don't think the reset index is needed here, we set the env variable to return the id as a column above</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1095/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

jasminerienecker: ah yes it seems extra logic was needed when we are tuning the hyperparameter
```
for exog in exogs:
  if isinstance(exog, str):
      hist_exog.append(exog)
  else:
      hist_exog.extend(exog)
```

jasminerienecker: yes removing it seems to still work well

jmoralez (approved):

jmoralez:
Thanks!

Issue #400, opened 2023-01-10
[BUG] cross_validation protections


kdgutier:
NeuralForecast `cross_validation` method breaks when `validation + test + lags > series length`.

It would be convenient to add a `Raise.Exception` protecting the user at the beginning of the method.

marcopeix:
Hello @kdgutier , can you provide a snippet of code, using the latest version, that replicates this issue? Thank you!

PR #1097, opened 2024-08-01
Fix (neuraforecast/core/cross-validation): Raise Exception when val_size+test_size is greater than min size of series.

marcopeix:
Right, if val_size+test_size is greater than the min size of a series, we show a warning, but the code still tries to run. In my test using the ETTm1 dataset, the following didn't fail immediately, but ran indefinitely: 

```python
nf.cross_validation(Y_df, val_size=2000000, test_size=2000000, n_windows=None)
```

I think it should raise an exception instead and stop running. I think this should solve #400


jmoralez:
I think the underlying issue is that a bunch of windows filled with zeros are used for training, so this would be fixed by #1059.

marcopeix:
> I think the underlying issue is that a bunch of windows filled with zeros are used for training, so this would be fixed by #1059.

Would it also solve #400 ?

jmoralez:
I don't know what that means by "breaks". Do you get an error if you specify those numbers to be just slightly above the minimum length?

marcopeix:
> I don't know what that means by "breaks". Do you get an error if you specify those numbers to be just slightly above the minimum length?

I get other errors, like `Exception: Time series is too short for training, consider setting a smaller input size or set start_padding_enabled=True`. You're right, the issue is not clear. I will ask for clarification and close this PR as it seems to introduce more problems than solutions.

PR #1129, opened 2024-08-29
hotfix: replace self.pred_len with self.h

marcopeix:
This fixes the attribute that does not exist when `channel_independence=1`.


elephaint (approved):
LGTM!

PR #1138, opened 2024-09-04
FIX: timemixer shapes mismatch and doc update

marcopeix:
Prevent shape mismatch in TimeMixer
Update docs: KAN supports exog feature, but currently it says that it doesn't support them


neuralforecast/models/timemixer.py
```diff
@@ -200,13 +201,13 @@ def __init__(self, seq_len, down_sampling_window, down_sampling_layers):
             [
                 nn.Sequential(
                     torch.nn.Linear(
-                        seq_len // (down_sampling_window ** (i + 1)),
-                        seq_len // (down_sampling_window**i),
+                        math.ceil(seq_len / (down_sampling_window ** (i + 1))),
```
carusyte, line 204: I tried using math.ceil on my end as well but it seems unable to solve all the edge cases, due to the variant output shape of previous conv / pooling modules. Currently I'm trying to work on a more stable solution by pre-calculating the output shape of those layers and pass them through down the line...

marcopeix: I guess we should stick to using the average as the down sampling method, since configuration of the model for the paper only uses that anyway. That will avoid the edge cases of shapes mismatch.

neuralforecast/models/timemixer.py
```diff
@@ -200,13 +201,13 @@ def __init__(self, seq_len, down_sampling_window, down_sampling_layers):
...
+                        math.ceil(seq_len / (down_sampling_window**i)),
                     ),
                     nn.GELU(),
                     torch.nn.Linear(
-                        seq_len // (down_sampling_window**i),
-                        seq_len // (down_sampling_window**i),
+                        math.ceil(seq_len // (down_sampling_window**i)),
```
elephaint, line 209: Here `math.ceil` isn't necessary. Above it is. Should we make it consistent? I.e. either use `//` in both cases or use `/` with `math.ceil`?

elephaint (changes requested):
Nice - minor detail to make it consistent

elephaint (approved):
Good work!

PR #1149, opened 2024-09-16
[FIX] CV Refit fails with non-standard column names

elephaint:
Fix for cross-validation with `refit=True` failing when using non-standard (id, time, target) names because they are not passed to the `.fit` method in `core.ipynb`


marcopeix (approved):
Good catch!

PR #1158, opened 2024-09-26
[FIX] Minor bugs

elephaint:
As #1023 is getting out of hand, splitting up the pieces.

- Replaces `.item()` in favor of `.detach().item()`
- Increase dof in Student-t to 3 to improve stability
- Improve speed of `_divide_no_nan`
- Remove StemGNN from multivariate evaluation in CI, it's horribly slow.
- Pad floats not int in ConstantPad1D (this errors when trying to compile a model)


jmoralez:
The purpose of `.item()` is converting a single element tensor to a python scalar. I'm guessing `detach` here is used to ensure the async execution of the GPU isn't blocked by that call, but wouldn't it be better to just prepend the detach call and keep the item? i.e. `Tensor.detach().item()`?

elephaint:
> The purpose of `.item()` is converting a single element tensor to a python scalar. I'm guessing `detach` here is used to ensure the async execution of the GPU isn't blocked by that call, but wouldn't it be better to just prepend the detach call and keep the item? i.e. `Tensor.detach().item()`?

Yeah you're right, I was lazy here, since the scalar didn't seem strictly necessary to create the log I dropped it, but happy to add it back

neuralforecast/common/_base_model.py
```diff
@@ -421,7 +421,7 @@ def on_validation_epoch_end(self):
         if self.val_size == 0:
             return
         losses = torch.stack(self.validation_step_outputs)
-        avg_loss = losses.mean().item()
+        avg_loss = losses.mean().detach()
```
jmoralez, line 424: missed this one

jmoralez (approved):

PR #1170, opened 2024-10-02
[FIX] MLPMultivariate incorrect static_exog parsing

elephaint:
`MLPMultivariate` parses `stat_exog` incorrectly, causing an error. This PR fixes that.

The updated example in the documentation of `MLPMultivariate` fails without this fix.


marcopeix (approved):
Looks good!

Issue #995, opened 2024-05-09
Conformal Prediction in NeuralForecast


valeman:
### Description

It would be great to have Conformal Prediction in NeuralForecast, similar to statsforecast and mlforecast.

PR #1171, opened 2024-10-03
[FEAT] Conformal Predictions in NeuralForecast

JQGoh:
# Rationale and Changes
* https://github.com/Nixtla/neuralforecast/issues/995
* Adopt the implementations in mlforecast for conformal predictions such as
  - https://github.com/Nixtla/mlforecast/pull/86
  - https://github.com/Nixtla/mlforecast/pull/97
* Add tutorial on conformal predictions

# Caveats
 * This does not support the dataframe of the type `SparkDataFrame`
 * Quantiled-type losses are not supported (do not conformalized various quantiled outputs)


elephaint:
This is great, let me know if I can help you with this

neuralforecast/utils.py
```diff
@@ -439,10 +442,124 @@ def augment_calendar_df(df, freq="H"):
...
+
+# %% ../nbs/utils.ipynb 32
+def add_conformal_distribution_intervals(
+    fcst_df: DataFrame,
+    cs_df: DataFrame,
+    model_names: List[str],
+    level: List[Union[int, float]],
+    cs_n_windows: int,
+    n_series: int,
+    horizon: int,
```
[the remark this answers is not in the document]
valeman: This is big 👍🚀

nbs/docs/tutorials/20_conformal_prediction.ipynb
```diff
@@ -0,0 +1,230 @@
...
+   "source": [
+    "# Conformal Prediction\n",
+    "> Tutorial on how to train neuralforecast models and obtain prediction intervals using the conformal prediction methods"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Conformal prediction intervals use cross-validation on a point forecaster model to generate the intervals. This means that no prior probabilities are needed, and the output is well-calibrated. No additional training is needed, and the model is treated as a black box. The approach is compatible with any model.\n",
```
elephaint, line 15: `Conformal prediction uses cross-validation on a model trained with a point loss function to generate prediction intervals. No additional training is needed, and the model is treated as a black box. The approach is compatible with any model.`

nbs/docs/tutorials/20_conformal_prediction.ipynb
```diff
@@ -0,0 +1,230 @@
...
+    "\n",
+    "In this notebook, we demonstrate how to obtain intervals using the conformal prediction methods."
```
elephaint, line 17: "In this notebook, we demonstrate how to obtain prediction intervals using conformal prediction."

nbs/docs/tutorials/20_conformal_prediction.ipynb
```diff
@@ -0,0 +1,230 @@
...
+    "os.environ['NIXTLA_ID_AS_COL'] = '1'"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Data\n",
+    "\n",
+    "We simply use AirPassengers dataset for the demostration of conformal predictions.\n"
```
elephaint, line 62: "We simply use the AirPassengers dataset for the demonstration of conformal prediction.

nbs/docs/tutorials/20_conformal_prediction.ipynb
```diff
@@ -0,0 +1,230 @@
...
+    "AirPassengersPanel_test = AirPassengersPanel[AirPassengersPanel['ds'] >= AirPassengersPanel['ds'].values[-12]].reset_index(drop=True)\n",
+    "AirPassengersPanel_test['y'] = np.nan\n",
+    "AirPassengersPanel_test['y_[lag12]'] = np.nan"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You can also create this directory structure with a spark dataframe using the following:"
```
elephaint, line 81: ? Is the Spark part necessary here? I think for most users it's easier to just show on a Pandas df

nbs/docs/tutorials/20_conformal_prediction.ipynb
```diff
@@ -0,0 +1,230 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Model training\n",
+    "\n",
+    "We now train a NHITS model on the above dataset. \n",
+    "It is worth noting that NeuralForecast currently does not support scaling when using this DataLoader. If you want to scale the timeseries this should be done before passing it in to the `fit` method."
```
elephaint, line 127: Maybe easier thus with a Pandas df

nbs/docs/tutorials/20_conformal_prediction.ipynb
```diff
@@ -0,0 +1,230 @@
...
+    "nf.fit(df=files_list, static_df=static, id_col='id_col')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Forecasting\n",
+    "\n",
+    "When working with large datasets, we need to provide a single DataFrame containing the input timesteps of all the timeseries for which wish to generate predictions. If we have future exogenous features, we should also include the future values of these features in the separate `futr_df` DataFrame. \n",
```
elephaint, line 157: for which we wish

nbs/docs/tutorials/20_conformal_prediction.ipynb
```diff
@@ -0,0 +1,230 @@
...
+    "predictions = nf.predict(df=pred_df, futr_df=futr_df, static_df=static)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| hide\n",
```
elephaint, line 183: I think this is still unfinished, right? 

I think the example would benefit from a plot showing the conformal intervals, ideally compared against a second setup where we just do DIstributionLoss("Normal") or something like that, displaying the width of the uncertainty intervals side-by-side (you can copy the code for such interval plots from a model page, e.g. BiTCN has it

nbs/losses.pytorch.ipynb
```diff
@@ -4361,6 +4361,46 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "c960d4db",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| export\n",
+    "UNSUPPORTED_LOSSES_CONFORMAL = (\n",
```
elephaint, line 4391: I don't think we need this variable. Point losses have `loss.outputsize_multiplier = 1`, and not-point losses have `loss.outputsize_multiplier > 1`. So i'd use that as check wherever you need it. This also generalizes over future new loss functions.

nbs/utils.ipynb
```diff
@@ -547,6 +556,159 @@
...
+    "    cs_df: DataFrame,\n",
+    "    model_names: List[str],\n",
+    "    level: List[Union[int, float]],\n",
+    "    cs_n_windows: int,\n",
+    "    n_series: int,\n",
+    "    horizon: int,\n",
+    ") -> DataFrame:\n",
+    "    \"\"\"\n",
+    "    Adds conformal intervals to a `fcst_df` based on conformal scores `cs_df`.\n",
+    "    `level` should be already sorted. This strategy creates forecasts paths\n",
```
elephaint, line 623: Do we enforce the sorting before calling `add_conformal_distribution_intervals`?

nbs/utils.ipynb
```diff
@@ -547,6 +556,159 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| export\n",
+    "\n",
+    "def add_conformal_error_intervals(\n",
+    "    fcst_df: DataFrame, \n",
```
elephaint, line 658: @jose-moralez I think this should take `DFType` instead of `DataFrame`, similar as to how MLForecast does it, right? 

See [here](https://github.com/Nixtla/mlforecast/blob/fa402075a7764a1374438261dc20259dee6ca19c/mlforecast/forecast.py#L77)

neuralforecast/core.py
```diff
@@ -69,6 +69,8 @@
     RMoK,
 )
 from .common._base_auto import BaseAuto, MockTrial
+from .utils import ConformalIntervals, get_conformal_method
+from .losses.pytorch import UNSUPPORTED_LOSSES_CONFORMAL
```
elephaint, line 73: See earlier comment, just check on `loss.outputsize_multiplier`

neuralforecast/core.py
```diff
@@ -483,6 +488,9 @@ def fit(
         ):
             raise Exception("Set val_size>0 if early stopping is enabled.")
 
+        self._cs_df: Optional[DataFrame] = None
```
elephaint, line 491: @jose-moralez same here, why aren't we using `DFType` here?

neuralforecast/core.py
```diff
@@ -648,10 +667,14 @@ def _get_needed_exog(self):
 
         return futr_exog | set(hist_exog)
 
-    def _get_model_names(self) -> List[str]:
+    def _get_model_names(self, use_conformal=False) -> List[str]:
         names: List[str] = []
         count_names = {"model": 0}
         for model in self.models:
+            # skip model for consideration of conformal prediction
+            if use_conformal and isinstance(model.loss, UNSUPPORTED_LOSSES_CONFORMAL):
```
elephaint, line 675: `if use_conformal and model.loss.outputsize_multiplier > 1`:

neuralforecast/core.py
```diff
@@ -943,6 +969,27 @@ def predict(
...
             _warn_id_as_idx()
             fcsts_df = fcsts_df.set_index(self.id_col)
+
+        # perform conformal predictions
+        if conformal_level is not None:
+            if self._cs_df is None or self.conformal_intervals is None:
+                warn_msg = "Please rerun the `fit` method passing a valid conformal_interval settings to compute conformity scores"
+                warnings.warn(warn_msg, UserWarning)
+            else:
+                level_ = sorted(conformal_level)
```
elephaint, line 979: Check, I guess the sorting is enforced here

elephaint (changes requested):
Awesome work! I did a first pass, I'll clone the branch tomorrow to take a deeper dive!

nbs/losses.pytorch.ipynb
elephaint: In your quick test the output should be the same as when using `MAE` (as it's the same loss). There's only a difference in the naming (when using `MAE` the output name is `NHITS`, when using `MQLoss([ ])` the output name is `NHITS-median`. But the amount of conformal predictions should be the same, only the column names should be different - I think the way you are suggesting right now.

neuralforecast/core.py
```diff
@@ -648,10 +667,14 @@ def _get_needed_exog(self):
 
         return futr_exog | set(hist_exog)
 
-    def _get_model_names(self) -> List[str]:
+    def _get_model_names(self, use_conformal=False) -> List[str]:
```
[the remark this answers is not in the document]
elephaint: Afaik conformal predictions just add prediction intervals over points. These points can be generated by a model trained using a point loss function that can also be a quantile loss with `q=80`. The interpretation of the results becomes a bit harder, but I'd be fine with having this functionality.

neuralforecast/utils.py
```diff
@@ -439,10 +442,130 @@ def augment_calendar_df(df, freq="H"):
...
+
+# %% ../nbs/utils.ipynb 31
+class ConformalIntervals:
+    """Class for storing conformal intervals metadata information."""
+
+    def __init__(
+        self,
+        n_windows: int = 2,
+        method: str = "conformal_distribution",
+        enable_quantiles: bool = False,
```
elephaint, line 461: Remove 'enable_quantiles' everywhere, the distinction doesn't make sense

neuralforecast/utils.py
```diff
@@ -439,10 +442,130 @@ def augment_calendar_df(df, freq="H"):
...
+        if n_windows < 2:
+            raise ValueError(
+                "You need at least two windows to compute conformal intervals"
+            )
+        allowed_methods = ["conformal_error", "conformal_distribution"]
+        if method not in allowed_methods:
+            raise ValueError(f"method must be one of {allowed_methods}")
+        self.n_windows = n_windows
+        self.method = method
+        self.enable_quantiles = enable_quantiles
```
elephaint, line 483: Should be removed

neuralforecast/utils.py
```diff
@@ -439,10 +442,130 @@ def augment_calendar_df(df, freq="H"):
...
+        scores = scores[:, :, :horizon]
+        mean = fcst_df[model].to_numpy().reshape(1, n_series, -1)
+        scores = np.vstack([mean - scores, mean + scores])
+        quantiles = np.quantile(
+            scores,
+            cuts,
+            axis=0,
+        )
+        quantiles = quantiles.reshape(len(cuts), -1).T
+        lo_cols = [f"{model}-conformal-lo-{lv}" for lv in reversed(level)]
```
elephaint, line 519: Naming should be without the `-conformal` tag, so that naming is identical to normal distribution outputs

neuralforecast/core.py
```diff
@@ -943,6 +972,30 @@ def predict(
...
+        # perform conformal predictions
+        if conformal_level is not None:
+            if self._cs_df is None or self.conformal_intervals is None:
+                warn_msg = "Please rerun the `fit` method passing a valid conformal_interval settings to compute conformity scores"
+                warnings.warn(warn_msg, UserWarning)
+            else:
+                level_ = sorted(conformal_level)
+                model_names = self._get_model_names(
+                    conformal=True,
+                    enable_quantiles=self.conformal_intervals.enable_quantiles,
```
elephaint, line 985: Remove `enable_quantiles`

elephaint (changes requested):
Great new stuff! I think most of the elements are there, I think on a more high-level I'm still a bit pondering whether this is the best way of including this in the API (I'll get back to that)

Couple of points for now:

- Please remove the `enable_quantiles` everywhere, I think it's unnecessary and the distinction between various point losses is arbitrary;
- Please remove the `-conformal` tag in the output names, this way the output names will be identical to normal DistributionLoss output names;
- `cross_validation` needs the option for conformal intervals too;
- The example needs a bit more work, for example below is a code snippet for creating a somewhat nicer plot

Example code in the tutorial (this already assumes the `-conformal` tag will be removed from the output name):

```
horizon = 12
input_size = 24

conformal_intervals = ConformalIntervals()

models = [NHITS(h=horizon, input_size=input_size, max_steps=100), NHITS(h=horizon, input_size=input_size, max_steps=100, loss=DistributionLoss("Normal", level=[90]))]
nf = NeuralForecast(models=models, freq='ME')
nf.fit(AirPassengersPanel_train, conformal_intervals=conformal_intervals)
```

```
fig, (ax1, ax2) = plt.subplots(2, 1, figsize = (20, 7))
plot_df = pd.concat([AirPassengersPanel_train, preds])

plot_df = plot_df[plot_df['unique_id']=='Airline1'].drop(['unique_id','trend','y_[lag12]'], axis=1).iloc[-50:]

ax1.plot(plot_df['ds'], plot_df['y'], c='black', label='True')
ax1.plot(plot_df['ds'], plot_df['NHITS1'], c='blue', label='median')
ax1.fill_between(x=plot_df['ds'][-12:], 
                 y1=plot_df['NHITS1-lo-90'][-12:].values,
                 y2=plot_df['NHITS1-hi-90'][-12:].values,
                 alpha=0.4, label='level 90')
ax1.set_title('AirPassengers Forecast', fontsize=18)
ax1.set_ylabel('Monthly Passengers', fontsize=15)
ax1.legend(prop={'size': 10})
ax1.grid()

ax2.plot(plot_df['ds'], plot_df['y'], c='black', label='True')
ax2.plot(plot_df['ds'], plot_df['NHITS'], c='blue', label='median')
ax2.fill_between(x=plot_df['ds'][-12:], 
                 y1=plot_df['NHITS-lo-90'][-12:].values,
                 y2=plot_df['NHITS-hi-90'][-12:].values,
                 alpha=0.4, label='level 90')
ax2.set_ylabel('Monthly Passengers', fontsize=15)
ax2.set_xlabel('Timestamp [t]', fontsize=15)
ax2.legend(prop={'size': 10})
ax2.grid()
```

nbs/core.ipynb
```diff
@@ -3204,6 +3321,147 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| hide\n",
+    "# test conformal prediction, method=conformal_distribution\n",
+    "\n",
+    "conformal_intervals = ConformalIntervals()\n",
+    "\n",
+    "models = []\n",
+    "for nf_model in [NHITS, RNN, StemGNN]:\n",
```
elephaint, line 3338: Please replace `StemGNN` with `TSMixer` everywhere where you are using `StemGNN` (it's another Multivariate model, but it's way faster than `StemGNN`, so this speeds up the tests)

nbs/core.ipynb
```diff
@@ -3204,6 +3321,147 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "0441e9d5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| hide\n",
+    "# test conformal prediction, method=conformal_distribution\n",
```
elephaint, line 3333: Love all the tests!

nbs/core.ipynb
```diff
@@ -3204,6 +3321,147 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d25b2cd2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| hide\n",
+    "# test conformal prediction are not applied for models with quantiled-related loss\n",
```
elephaint, line 3409: This test isn't necessary when removing `enable_quantiles` argument. It's fine to apply conformal to any quantile loss.

nbs/core.ipynb
```diff
@@ -989,6 +1014,29 @@
     "        if isinstance(fcsts_df, pd.DataFrame) and _id_as_idx():\n",
     "            _warn_id_as_idx()\n",
     "            fcsts_df = fcsts_df.set_index(self.id_col)\n",
+    "\n",
+    "        # perform conformal predictions\n",
+    "        if conformal_level is not None:\n",
```
elephaint, line 1019: I think we should just call this `level`, that will also generalize it when we want to use the argument to generate arbitrary levels with DistributionLoss functions.

nbs/core.ipynb
```diff
@@ -989,6 +1014,29 @@
...
+    "            if self._cs_df is None or self.conformal_intervals is None:\n",
+    "                warn_msg = (\n",
+    "                    'Please rerun the `fit` method passing a valid conformal_interval settings to compute conformity scores'\n",
+    "                )\n",
+    "                warnings.warn(warn_msg, UserWarning)\n",
```
elephaint, line 1024: I think we should raise an Exception here - if a user passes a level but it's not possible compute that level an Exception feels more appropriate.

nbs/core.ipynb
```diff
@@ -3204,6 +3321,147 @@
...
+    "       'NHITS-lo-80-conformal-lo-90', 'NHITS-lo-80-conformal-lo-50',\n",
+    "       'NHITS-lo-80-conformal-lo-10', 'NHITS-lo-80-conformal-hi-10',\n",
+    "       'NHITS-lo-80-conformal-hi-50', 'NHITS-lo-80-conformal-hi-90',\n",
+    "       'NHITS-hi-80-conformal-lo-90', 'NHITS-hi-80-conformal-lo-50',\n",
+    "       'NHITS-hi-80-conformal-lo-10', 'NHITS-hi-80-conformal-hi-10',\n",
+    "       'NHITS-hi-80-conformal-hi-50', 'NHITS-hi-80-conformal-hi-90'\n",
+    "]\n",
+    "\n",
+    "assert all([col in preds.columns for col in pred_cols])\n"
+   ]
```
elephaint, line 3464: How do you deal with the case when non-point losses are used in combination with conformal intervals?

elephaint:
@JQGoh I pushed most of the suggested fixes already, saving you some time.

I think we still need:
- [ ] Cross validation prediction intervals
- [ ] Some more tests (I need to think about the relevant tests)

Other than that I'm really happy with what you made and I think we're almost there.

nbs/core.ipynb
elephaint: I think we need more tests, a.o. what exceptions/warnings are raised when we provide combinations of loss functions / prediction intervals that don't work together. But I believe we already capture most of these.

elephaint: I'll resolve this conversation and others to continue [from here](https://github.com/Nixtla/neuralforecast/pull/1171#issuecomment-2399549031)

neuralforecast/utils.py
```diff
@@ -439,10 +442,124 @@ def augment_calendar_df(df, freq="H"):
...
+
+    def __init__(
+        self,
+        n_windows: int = 2,
+        method: str = "conformal_distribution",
+    ):
+        """
+        n_windows : int
+            Number of windows to evaluate.
+        method : str, default is conformal_distribution
```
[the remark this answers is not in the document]
elephaint: I think this is a good decision, albeit I renamed it to `prediction_intervals` to conform to MLForecast API and allow generalization to non-conformal methods down the road.

elephaint:
> > @JQGoh I pushed most of the suggested fixes already, saving us some time.
> > I think we still need:
> > 
> > * [ ]   Cross validation prediction intervals (we should be able to follow MLForecast example)
> > * [ ]   Some more tests (I need to think about the relevant tests)
> > * [ ]   Spark DataFrame support (?)
> > 
> > Other than that I'm really happy with what you made and I think we're almost there.
> 
> @elephaint Thanks for your help with the revision. I will think about the mentioned items. By the way, it appears that for now we only introduce conformal predictions to point loss, but not on the quantiled outputs? If that is the case, think we better mention this in the channel as previously I said that I wanted to introduce an optional parameter that supports this.

Correct - users can still get prediction intervals over an arbitrary quantile output by fitting a model with e.g. `QuantileLoss(q=0.8)`. Maybe we change this in the future, but it feels a bit too meta for now to include prediction intervals over prediction intervals.

nbs/core.ipynb
```diff
@@ -3204,6 +3323,112 @@
...
+    "# refit=False, no conformal predictions outputs\n",
+    "nf = NeuralForecast(models=[NHITS(h=12, input_size=24, max_steps=1)], freq='M')\n",
+    "cv1 = nf.cross_validation(\n",
+    "    AirPassengersPanel_train, \n",
+    "    prediction_intervals=prediction_intervals,\n",
+    "    level=[30]\n",
+    "    )\n",
+    "assert not any([col in cv1.columns for col in ['NHITS-lo-30', 'NHITS-hi-30']])\n",
+    "\n",
+    "# refit=True, we have conformal predictions outputs\n",
```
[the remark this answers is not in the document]
elephaint: Thanks for these changes! I'll have a look

elephaint:
@JQGoh Thanks again! I added some protections and moved the conformity score calculation to within the parts where we are able to do it. I think it will work like this, only the situation with a stored dataset (`df=None`) I haven't fully covered, although I'm not sure that's a major issue.

Let me know what you think

elephaint (approved):

elephaint:
Ok, thanks again @JQGoh - I'm happy to merge

neuralforecast/utils.py
```diff
@@ -439,10 +443,126 @@ def augment_calendar_df(df, freq="H"):
...
+    Adds conformal intervals to a `fcst_df` based on conformal scores `cs_df`.
+    `level` should be already sorted. This strategy creates forecasts paths
+    based on errors and calculate quantiles using those paths.
+    """
+    fcst_df = ufp.copy_if_pandas(fcst_df, deep=False)
+    alphas = [100 - lv for lv in level]
+    cuts = [alpha / 200 for alpha in reversed(alphas)]
+    cuts.extend(1 - alpha / 200 for alpha in alphas)
+    for model in model_names:
+        scores = cs_df[model].to_numpy().reshape(cs_n_windows, n_series, horizon)
```
jmoralez, line 505: This works in mlforecast because the outputs are sorted by window and then series, but in neural they're sorted by series first, so this has to be:
```python
scores = cs_df[model].to_numpy().reshape(n_series, cs_n_windows, horizon)
scores = scores.transpose(1, 0, 2)
```

neuralforecast/core.py
```diff
@@ -24,7 +24,7 @@
     LocalRobustScaler,
     LocalStandardScaler,
 )
-from utilsforecast.compat import DataFrame, Series, pl_DataFrame, pl_Series
+from utilsforecast.compat import DataFrame, DFType, Series, pl_DataFrame, pl_Series
```
jmoralez, line 27: This was added in utilsforecast 0.2.3, so it has to be updated in the settings.ini

neuralforecast/utils.py
elephaint: Added in fix PR

neuralforecast/core.py
elephaint: Added in fix PR

AzulGarza:
hey @JQGoh! we would like to talk with you. is there an email address where we can contact you? if you prefer, feel free to send me a message at `azul@nixtla.io`.

PR #1175, opened 2024-10-10
FEAT - Set activation function in GRN of TFT

marcopeix:
Right now, the docstring specifies a choice of activation function, but if used, it throws an error.

This PR adds the option of specifying the activation function in the GRN component. We also remove the `shared_weights` parameter as it was unused.

This code now runs without failing:

```python
import pandas as pd
import numpy as np
from neuralforecast.models import TFT

AirPassengersPanel['month']=AirPassengersPanel.ds.dt.month
Y_train_df = AirPassengersPanel[AirPassengersPanel.ds<AirPassengersPanel['ds'].values[-12]] # 132 train
Y_test_df = AirPassengersPanel[AirPassengersPanel.ds>=AirPassengersPanel['ds'].values[-12]].reset_index(drop=True) # 12 test

nf = NeuralForecast(
    models=[TFT(h=12, input_size=48,
                hidden_size=20,
                grn_activation='ELU',
                loss=DistributionLoss(distribution='StudentT', level=[80, 90]),
                learning_rate=0.005,
                stat_exog_list=['airline1'],
                futr_exog_list=['y_[lag12]','month'],
                hist_exog_list=['trend'],
                max_steps=300,
                val_check_steps=10,
                early_stop_patience_steps=10,
                scaler_type='robust',
                windows_batch_size=None,
                enable_progress_bar=True),
    ],
    freq='M'
)
nf.fit(df=Y_train_df, static_df=AirPassengersStatic, val_size=12)
```


elephaint (approved):
LGTM!

PR #1191, opened 2024-11-03
DynamicNBEATs model

tblume1992:
By default NBEATs only utilizes a polynomial basis for the trend component. This PR adds support for several different basis functions for the trend and adds a handler for these functions in a new 'DynamicNBEATs' method.


elephaint:
Hi, thanks for your contribution, this is really cool! 

Could you please follow the [contributing guidelines](https://github.com/Nixtla/neuralforecast/blob/main/CONTRIBUTING.md)? Specifically, we need you to build the library so that the correct python files are generated.

tblume1992:
1. hey @elephaint I was chatting with @marcopeix about this awhile back. Unfortunately I don't have time for a PR but he wanted me to be the author of it. All the code was just in a notebook I threw into the PR. 

2. You can definitely do that and that is basically what I did. 95% of this code is just the base NBEATs code. I think though that this change would alter the 'auto' model pretty significantly since we would want to search for the basis functions AND the number of functions. So we may want to think of it as a different model.

marcopeix (approved):

elephaint (dismissed):
LGTM!

marcopeix:
I'm happy with this version of NBEATS now. Thank you @tblume1992 for your patience! We'll adjust according to the review comments!

marcopeix (approved):

Antoine-Schwartz:
Hello,

First of all great job and thanks for this improvement! 
I was wondering if these changes could be applied in the same way to NBEATSx?

PR #1198, opened 2024-11-12
feat: Deprecate activation functions for GRU

marcopeix:
Currently, there's a parameter for the encoder activation function, but it is not used. 

PyTorch has frozen the activation function in GRU, meaning that we cannot change this. This PR deprecates this argument and we'll remove it in a future version.


neuralforecast/models/gru.py
```diff
@@ -199,6 +218,7 @@ def forward(self, windows_batch):
         hidden_state, _ = self.hist_encoder(
             encoder_input
         )  # [B, seq_len, rnn_hidden_state]
+        hidden_state = self.encoder_activation(hidden_state)
```
elephaint, line 221: The encoder activation should not be set in .forward here, but in the definition of self.hist_encoder - see the arguments supported by GRU

elephaint (changes requested):
Encoder activation should be in the definition of GRU, not in the forward method

neuralforecast/models/gru.py
marcopeix: From the [documentation](https://pytorch.org/docs/stable/generated/torch.nn.GRU.html), there is no parameter for specifying an activation function, which is why I add it in the forward method.

marcopeix: Actually, maybe we should just close this PR and the related issue, and say that it's not possible to change the activation functions inside of GRU. 

I'd say we have two options:

1. Remove the parameter and use the default activation from PyTorch, which they froze and we can't modify.
2. Implement a custom GRU class where we can change the activation function (not sure if it's worth it, especially since it's not that performant for forecasting).

What do you think?

elephaint (changes requested):
I think we should raise a deprecation message if we deprecate the argument, and keep the argument.

neuralforecast/models/gru.py
```diff
@@ -71,6 +73,7 @@ def __init__(
         inference_input_size: int = -1,
         encoder_n_layers: int = 2,
         encoder_hidden_size: int = 200,
+        encoder_activation: str = "tanh",
```
elephaint, line 76: This should now be:

`encoder_activation: Optional[str] = None`

The default value is now None, so that the deprecation message doesn't show by default. Whenever somebody uses the old input "tanh", the deprecation message is shown.

The way it currently goes is that the deprecation warning is always shown, regardless of whether the argument is actively used by someone.

elephaint (changes requested):

neuralforecast/models/gru.py
marcopeix: Done

nbs/models.gru.ipynb
```diff
@@ -101,7 +111,7 @@
     "    `inference_input_size`: int, maximum sequence length for truncated inference. Default -1 uses all history.<br>\n",
     "    `encoder_n_layers`: int=2, number of layers for the GRU.<br>\n",
     "    `encoder_hidden_size`: int=200, units for the GRU's hidden state size.<br>\n",
-    "    `encoder_activation`: str=`tanh`, type of GRU activation from `tanh` or `relu`.<br>\n",
+    "    `encoder_activation`: Optinal[str]=None, Deprecated. Activation function in GRU is frozen in PyTorch.<br>\n",
```
elephaint, line 114: typo

elephaint (approved):
Non-blocking typo

PR #1220, opened 2024-11-27
breaking: remove deprecated behavior

jmoralez:
Removes the following behavior, which has been deprecated for about a year:

* Providing the id as the dataframe index, it has to be a column. Also the outputs from the methods will now always have the id as a column.
* Providing `sort_df` to `fit` and `cross_validation`, we always check if it's sorted.
* Providing `num_workers_loader` to the models constructor, those should now be provided through `dataloader_kwargs`.

Also updates the documentation to remove usages of this behavior, docs that use `datasetsforecast.losses` and uses `utilsforecast.plotting.plot_series` instead of `StatsForecast.plot`


nbs/docs/getting-started/01_introduction.ipynb
```diff
@@ -102,45 +102,74 @@
...
    "execution_count": null,
-   "id": "401a785d",
+   "id": "ea728145-d6cb-4eab-888a-6033b3388953",
    "metadata": {},
    "outputs": [],
    "source": [
-    "#| eval: false\n",
+    "import logging\n",
+    "\n",
     "import numpy as np\n",
```
elephaint, line 111: Non-blocking we could get rid of numpy, pandas  and IPython imports here too

nbs/docs/getting-started/05_datarequirements.ipynb
```diff
@@ -80,13 +80,15 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "100%|██████████| 1.76M/1.76M [00:00<00:00, 5.55MiB/s]\n",
-      "INFO:datasetsforecast.utils:Successfully downloaded M3C.xls, 1757696, bytes.\n"
+      "100%|█████████████████████████████████████████████████████████████████████████████████████████████| 1.76M/1.76M [00:00<00:00, 25.2MiB/s]\n",
```
elephaint, line 83: Non-blocking but nice to remove this cell output

elephaint (changes requested):
Mainly non-blocking comments, but the following notebooks still have `os.environ['NIXTLA_ID_AS_COL']` in it:

- 02_cross_validation
- 10_distributed
- 12_using_mlflow
- 19_large_datasets
- 20_conformal_prediction

nbs/core.ipynb
```diff
@@ -1869,46 +1807,6 @@
...
-    "models = [\n",
-    "    NHITS(h=12, input_size=12, max_steps=1)\n",
-    "]\n",
-    "nf = NeuralForecast(models=models, freq='M')\n",
-    "with warnings.catch_warnings(record=True) as issued_warnings:\n",
-    "    warnings.simplefilter('always', category=FutureWarning)\n",
-    "    nf.fit(df=df_with_idx)    \n",
-    "    nf.predict()\n",
-    "    nf.predict_insample()\n",
-    "    nf.cross_validation(df=df_with_idx)\n",
```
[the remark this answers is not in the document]
jmoralez: I had missed that one, thanks!

nbs/docs/capabilities/05_predictInsample.ipynb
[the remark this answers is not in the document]
jmoralez: Yeah that's because at first we don't have enough samples for the input size so they look like that.

nbs/docs/capabilities/06_save_load_models.ipynb
```diff
@@ -398,107 +506,31 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| hide\n",
+    "pd.testing.assert_frame_equal(Y_hat_df, Y_hat_df2)"
```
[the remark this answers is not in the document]
jmoralez: Fixed, thanks!

nbs/docs/getting-started/01_introduction.ipynb
jmoralez: Removed

nbs/docs/getting-started/05_datarequirements.ipynb
jmoralez: Removed

jmoralez:
I removed the uses of datasetsforecast.losses, the ones that set the env variable wouldn't be broken but it can be confusing. We still have to update all notebooks which have a reset_index in them but I wanted to finish this within my lifetime so I didn't do it here but I'll update this PR when I get time to run those as well.

elephaint (approved):
Thanks!

PR #1230, opened 2024-12-17
Custom rnn layers for TFT

Yanam24:
Custom TFT
- Add the possibility to chose between LSTM and GRU layers 
- Add the possibility to stack multiple RNN layers 
- Chose to initialize all the RNN states with the same or different values computed from static covariates


marcopeix (approved):
LGTM!

PR #1233, opened 2024-12-18
Feat: We added the horizon weighing to the distribution losses, simil…

mwamsojo:
We suggest modifying the distribution loss class to include the weighting option for distribution losses (to be similar to point losses)

- We pass the horizon weight vector when initializing the losses
- We add a function to check and apply the weights according to the mask


marcopeix (approved):
LGTM

PR #1243, opened 2025-01-03
Update models.nhits.ipynb

k3ybladewielder:
Adjust the freq 'M' to 'ME'. 'M' is deprecated.


elephaint:
Thank you for pointing this out to us! We decided to address it for all models in #1274

Issue #1259, opened 2025-02-02
[Core|Common] TypeError: Got unsupported ScalarType BFloat16


carusyte:
### What happened + What you expected to happen

Nowadays BFloat16 support are getting prevalent for ML libs and hardware. Much appreciated if bfloat16 support can be applied in Neuralforecast as well.

```
...
831   File "/home/kemove/git/neuralforecast/neuralforecast/core.py", line 1345, in predict_insample
832     model_fcsts = model.predict(trimmed_dataset, step_size=step_size)
833                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
834   File "/home/kemove/git/neuralforecast/neuralforecast/common/_base_multivariate.py", line 607, in predict
835     fcsts = torch.vstack(fcsts).numpy()
836             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
837 TypeError: Got unsupported ScalarType BFloat16
...
```

### Versions / Dependencies

Neuralforecast: 2.0.0

### Reproduction script

N/A

### Issue Severity

None

elephaint:
Hey, do you have a piece of code so that I reproduce the error?

carusyte:
Sure. Generally the error is thrown when I ran this "minimal" reproducer on my workstation with CPU / GPU:

```python
import pandas as pd
import numpy as np
import logging
import torch
from neuralforecast.core import NeuralForecast
from neuralforecast.models import TSMixerx
 
# Prep dummy data
start_date = "2000-01-01"
end_date = "2025-02-01"
date_range = pd.date_range(start=start_date, end=end_date, freq="B")
np.random.seed(0)
df = pd.DataFrame(
    {
        "unique_id": "dummy",
        "ds": date_range,
        "y": np.random.randn(len(date_range)),
        "val1": np.random.randn(len(date_range)),
        "val2": np.random.rand(len(date_range)),
    }
)
logging.getLogger("pytorch_lightning").setLevel(logging.ERROR)
torch.set_float32_matmul_precision("medium")
horizon = 10
input_size = 20
models = [
    TSMixerx(
        h=horizon,
        input_size=input_size,
        n_series=1,
        revin=False,
        max_steps=100,
        random_seed=0,
        hist_exog_list=["val1", "val2"],
        precision="bf16-mixed",
    ),
]
nf = NeuralForecast(
    models=models,
    freq="B",
    local_scaler_type="robust",
[...]

PR #1260, opened 2025-02-02
add bfloat16 support in _base_multivariate

carusyte:
This PR is a proposal to address #1259 

Ref:
https://github.com/pytorch/pytorch/issues/109873
https://github.com/pytorch/pytorch/issues/110285

### DIFF
--- a/neuralforecast/common/_base_multivariate.py
+++ b/neuralforecast/common/_base_multivariate.py
@@ -5,6 +5,7 @@ __all__ = ['BaseMultivariate']
 
 # %% ../../nbs/common.base_multivariate.ipynb 5
 import numpy as np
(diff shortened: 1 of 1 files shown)


elephaint:
Thanks for your contribution!

Please follow [our contributing guide](https://github.com/Nixtla/neuralforecast/blob/main/CONTRIBUTING.md); we use `nbdev` so direct changes to .py files are generally not allowed. Also we'd have to incorporate this for all base* classes. But first, I think it would be good to understand the issue better with a piece of code I can run (see my post in the issue).

elephaint:
Closed in favour of #1265, which generalizes the fix and removes the additional dependency

PR #1263, opened 2025-02-04
Proposing a means to embed learning rate finder (lr_find) before fitting multivariate models

carusyte:
idea: https://github.com/Nixtla/neuralforecast/discussions/1262

### DIFF
--- a/neuralforecast/common/_base_model.py
+++ b/neuralforecast/common/_base_model.py
@@ -16,6 +16,7 @@ import numpy as np
 import torch
 import torch.nn as nn
 import pytorch_lightning as pl
+from pytorch_lightning.tuner.tuning import Tuner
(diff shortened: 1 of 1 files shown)


elephaint:
Thanks for your contribution;

Please follow [our contributing guide](https://github.com/Nixtla/neuralforecast/blob/main/CONTRIBUTING.md); we use `nbdev` so direct changes to .py files are generally not allowed. 

I like the idea as it seems simple and straightforward, can you provide a piece of code that showcases the functionality?

marcopeix:
No need of this PR, we can do:
```python
from pytorch_lightning.callbacks import LearningRateFinder
models = [NHITS(h=12, input_size=24, max_steps=10, learning_rate=1e-3, callbacks=[LearningRateFinder()])]
```

PR #1267, opened 2025-02-14
FEAT: TimeXer

marcopeix:
Add TimeXer to neuralforecast. This is a transformer-based model that supports future exogenous features, and seems to perform very well in long-horizon forecasting according to [here](https://github.com/thuml/Time-Series-Library).

Original code implementation: https://github.com/thuml/TimeXer/tree/main
Paper: https://arxiv.org/abs/2402.19072

**Note**: The encoder and embedding are specific to TimeXer, that's why I am not reusing the existing common module from other models.


nbs/common.modules.ipynb
```diff
@@ -409,40 +410,93 @@
    "source": [
     "#| export\n",
     "class AttentionLayer(nn.Module):\n",
-    "    def __init__(self, attention, hidden_size, n_head, d_keys=None,\n",
+    "    def __init__(self, attention, hidden_size, n_heads, d_keys=None,\n",
```
elephaint, line 413: If we're consolidating these layers here from various Transformer models, why not just use Pytorch's `nn.MultiheadAttention` in each of these separate models? That would allow us to remove all these common modules, too. 

Note that I think that also incorporates causal masking, so the triangular causal mask may not be necessary either.

We could split it into another PR though, although it might be a small change.

nbs/sidebar.yml
```diff
@@ -50,6 +50,7 @@ website:
           - models.timemixer.ipynb
           - models.timellm.ipynb
           - models.timesnet.ipynb
+          - models.timexer.ipynb
```
elephaint, line 53: Also update `mint.json`

elephaint (changes requested):
LGTM, maybe the refactoring in a separate PR?

We can also do it separately though, it's non-blocking imho.

Small detail re. `mint.json`. 

Does it (somewhat) reproduce the paper results?

nbs/sidebar.yml
marcopeix: Done

elephaint (approved):
LGTM!

marcopeix:
> LGTM, maybe the refactoring in a separate PR?
> 
> We can also do it separately though, it's non-blocking imho.
> 
> Small detail re. `mint.json`.
> 
> Does it (somewhat) reproduce the paper results?

Here are the results for ETTm1:

**Paper (avg for 4 horizons)**
MAE: 0.397
MSE: 0.382

**neuralforecast (avg for 4 horizons)**
MAE: 0.423
MSE: 0.445

Although I was training for 1000 step with early stopping on all horizons and they all reached 1000 steps, so maybe training for longer would give closer scores

PR #1271, opened 2025-02-21
FIX:  Insample predictions with multivariate model

marcopeix:
Currently, insample predictions do not work for multivariate model; they miss `input_size` values. This is likely due to padding not being applied for early timesteps when creating windows for multivariate models.

This fix allows padding, just like in `BaseWindows`. This fixes the problem of insample predictions with multivariate models.

I tested the performance of some multivariate models, and padding does not impact performance.

I also added a multivariate model to the insample predictions tests.


nbs/common.base_multivariate.ipynb
```diff
@@ -234,9 +234,13 @@
     "        elif step in ['predict', 'val']:\n",
     "\n",
     "            if step == 'predict':\n",
+    "                initial_input = temporal.shape[-1] - self.test_size\n",
+    "                if initial_input <= self.input_size: # There is not enough data to predict first timestamp\n",
```
elephaint, line 238: Maybe a unit test to verify that this works as intended?

marcopeix:
Insample predictions with multivariate models now works with the merge of #1023 . Closing as this PR is not necessary anymore.

Issue #1273, opened 2025-02-23
[Model: TFT] StaticCovariateEncoder fails with empty tensor concatenation error


emhedlin:
## The issue
When trying to use the TFT model with static features, the model fails with a `RuntimeError: torch.cat(): expected a non-empty list of Tensors`. 

The model failed with real world data I was using, so I tried with generate_series() data and it failed in the same way.

### Versions / Dependencies

neuralforecast version: `2.0.1`
python version: `3.10.12`
os: `Pop Os`

### Reproduction script

```
from neuralforecast.utils import generate_series
from neuralforecast.auto import TFT

train, static = generate_series(n_series = 4, freq='D', n_static_features=4)

model = [
    TFT(
        h=5,
        loss=MAE(),
        futr_exog_list=[],
        hist_exog_list=[],
        stat_exog_list=['static_0', 'static_1', 'static_2', 'static_3'],
        input_size=10,
        hidden_size = 10,
        n_head=5,
        max_steps=1500,
        enable_progress_bar=True,
    )
]

nf = NeuralForecast(models=model, freq='D')
nf.fit(df=train, static_df=static)
```

### Issue Severity

None

emhedlin:
Training still fails when I:

- Omit or change the `futr_exog_list` / `hist_exog_list` args to None
- Include known and unknown regressors:
```
from neuralforecast.utils import generate_series
from neuralforecast.auto import TFT

train, static = generate_series(
    n_series = 4, 
    freq='D', 
    n_static_features=4,
    n_temporal_features=4
)

unknown = ['temporal_0', 'temporal_1']
known = ['temporal_2', 'temporal_3']

model = [
    TFT(
        h=5,
        loss=MAE(),
        futr_exog_list=known,
        hist_exog_list=unknown,
        stat_exog_list=['static_0', 'static_1', 'static_2', 'static_3'],
        input_size=10,
        hidden_size = 10,
        n_head=5,
        max_steps=1500,
        enable_progress_bar=True,
    )
]

nf = NeuralForecast(models=model, freq='D')
nf.fit(df=train, static_df=static)
```


For transparency, I have known and unknown regressors with my real world data, and everything works perfectly as long as I specify no static regressors (`stat_exog_list = []` or `stat_exog_list = None`). As soon as I supply static regressors, I get the RuntimeError: `RuntimeError: torch.cat(): expected a non-empty list of Tensors`.

emhedlin:
Thanks @marcopeix , this works 👍

Issue #1268, opened 2025-02-17
[Core]: training with val_size > 0 does not work: maximum size for tensor at dimension _ is _ but size is _


ischurov:
### What happened + What you expected to happen

I am trying to fit `VanillaTransformer` model on a time series with `max_horizon = 2`. If I try to turn on the validation during training (i.e. pass `val_size > 0` in `.fit()`), I get an error that typically looks like 

```
...
  File "<...>/site-packages/neuralforecast/common/_base_windows.py", line 248, in _create_windows
    windows = temporal.unfold(
              ^^^^^^^^^^^^^^^^
RuntimeError: maximum size for tensor at dimension 2 is 5 but size is 6
```

It seem to be core issue, as it is also reproduced with `TFT` instead of `VanillaTransformer`.

Here are full logs:

[nixtla_testcase.log](https://github.com/user-attachments/files/18827831/nixtla_testcase.log)

### Versions / Dependencies

```
pandas == 2.2.3
neuralforecast == 2.0.1
```

Also reproduces at current development version of neuralforecast (3f4a837a0f61e65d831d4119769d9da33caa020f).

### Reproduction script

This is a self-contained reproduction script. One can run it with `uv run <script_name.py>`.

```
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "pandas == 2.2.3",
#   "neuralforecast == 2.0.1"
# ]
# ///

# Also reproduces with
# "neuralforecast @ git+https://github.com/Nixtla/neuralforecast@3f4a837a0f61e65d831d4119769d9da33caa020f",

import neuralforecast
import pandas as pd
from neuralforecast import NeuralForecast
from neuralforecast.models import TFT, VanillaTransformer

test_data = pd.DataFrame(
    {
        "ds": pd.date_range(
            start="2023-01-31",
            end="2025-01-31",
            freq="ME",
            inclusive="right",
        ),
        "unique_id": "series_1",
        "y": 0,
    }
)
print(f"{neuralforecast.__version__=}")
assert test_data.shape == (24, 3)
# we have 24 time steps, so there should be enough data

model = VanillaTransformer(h=2, input_size=4)

forecaster = NeuralForecast(models=[model], freq="M")

forecaster.fit(df=test_data, val_size=1)
```

### Issue Severity

Medium: It is a significant difficulty but I can work around it.

marcopeix:
[...]

PR #1275, opened 2025-02-25
FIX: Minor bug fix in TFT and a nicer error message for fitting with the wrong val_size

marcopeix:
Issue 1: #1273 
- The default value of `rnn_type` in TFT was in uppercase. This caused errors for the StaticCovariateEncoder. This PR simply ensures it's in lowercase, which fixes the issue.

Issue 2: #1268 
- Currently, there is no protection in setting `val_size` < `h`, which causes errors when fitting a model. This PR adds a protection and a nicer message in those situations.


elephaint (approved):
Good catch! LGTM

PR #1285, opened 2025-03-09
[FIX] Recurrent predictions

elephaint:
When setting `recurrent=True` in recursive models (LSTM, GRU, RNN) there is a bug that ignores the latest available ground truth in the forecast (i.e., off-by-one error). This makes forecasts slightly worse, thus fixing the bug improves forecast quality when making forecasts with `recursive=True`.


marcopeix (approved):
Looks good!

Issue #1080, opened 2024-07-25
Specify Ray Scheduler / BOHB Support


samuel-gerstein:
### Description

Currently, there isn’t a way to specify a ray scheduler, neuralforecast just uses the default FIFO one. If you want to use a sampling algorithm like BOHB, you need to use a different ray scheduler than the default one.

Issue #634, opened 2023-06-08
[Auto] Option to manage hyperparameter tuning with time budget


cchallu:
### Description

Tune's library `Tuner` class can stop hyperparameter search based on execution time (using `time_budget_s`) instead of the number of configurations (currently specified with `num_samples`). See https://docs.ray.io/en/latest/tune/key-concepts.html for more details.

**Instructions**

In the `base_auto` class, modify:

```python
tune_config=tune.TuneConfig(
    metric="loss",
    mode="min",
    num_samples=num_samples, 
    search_alg=search_alg
),
```

with 

```python
tune_config=tune.TuneConfig(
    metric="loss",
    mode="min",
    num_samples=num_samples,
    time_budget_s=time_budget,
    search_alg=search_alg
),
```

It will also require adding the `time_budget` hyperparameter (defaulted to `None`) to the `base_class` and intermediate functions (analogous to `num_samples`) and adding an Exception if `time_budget!=None` and `num_samples!=1`.

To contribute, follow the instructions at https://github.com/Nixtla/neuralforecast/blob/main/CONTRIBUTING.md.

Thanks!

PR #1286, opened 2025-03-11
[FEAT] Option to customize the hyperparameter tuning behavior of Ray

JQGoh:
Main objective is to provide a way to customize the hyperparameter tuning behavior of Ray. See issues
 * https://github.com/Nixtla/neuralforecast/issues/634 
 * https://github.com/Nixtla/neuralforecast/issues/1080


marcopeix:
Hi @JQGoh thanks a lot for the PR! I have few comments and questions.

- First of all, I think it's great that you're tackling this and we're definitely on the right track.
- Running the tutorial notebook on my end, I see that the predictions are exactly the same, whether we use the BHOB scheduler or not. Is that expected? It might be that AirPassengers is too simple, but I wanna make sure you see the same (in which case, we have to modify the conclusion of the tutorial)
- Can we include a time budget? I see you tagged that issue, but I don't see it in the tutorial.
- Can you merge main into the branch please?

Thanks a lot for your work!

marcopeix:
@JQGoh I tried some approaches on my end to avoid using `deepcopy`, but nothing worked. I don't think there's a simple solution to it. For now, I would keep it as is, and make sure that the functionality works, and the tutorial is adjusted. We can always think of a better approach later on. I believe the advantages outweigh the drawbacks here.

marcopeix:
> > @JQGoh I tried some approaches on my end to avoid using `deepcopy`, but nothing worked. I don't think there's a simple solution to it. For now, I would keep it as is, and make sure that the functionality works, and the tutorial is adjusted. We can always think of a better approach later on. I believe the advantages outweigh the drawbacks here.
> 
> @marcopeix Is there any plan to do a release soon? I try to wrap this up by end of the week if possible

I want to merge a couple of PRs before making a release (this one included), so no rush!

nbs/core.ipynb
```diff
@@ -551,7 +551,14 @@
...
     "            if prediction_intervals is not None:\n",
     "                self.prediction_intervals = prediction_intervals\n",
-    "                self._cs_df = self._conformity_scores(\n",
+    "\n",
+    "                # copy of model to prevent error during hyperparmeter tuning with Ray\n",
+    "                # ValueError: You passed a `param_space` parameter to `Tuner(...)` with unresolved parameters,\n",
+    "                #  but the search algorithm was already instantiated with a search space. \n",
+    "                # Make sure that `config` does not contain any more parameter definitions \n",
+    "                # - include them in the search algorithm's search space if necessary.\n",
+    "                model_copy = deepcopy(self)\n",
```
elephaint, line 560: @JQGoh for determining the conformity scores, isn't it sufficient to temporarily change some parameters of the model so that it doesn't give the error?

marcopeix:
Fixed by #1531

PR #1274, opened 2025-02-25
[DOCS] Update documentation

elephaint:
- Updates deprecated frequencies in model examples
- Updates documentation, using utilsforecast.losses instead of nf
- Re-run examples using latest NF version to make sure examples still run and conclusions hold. If not, example adjusted.
- Fixes a number of dead links
- Rewrite of the 'adding model' tutorial to reflect removal of Base* classes
- Small fix in use of `MQLoss` / `HuberMQLoss`
- Removed a warning message that's not necessary.

PR #1288, opened 2025-03-15
Update README.md

crishna0401:
In Quick start freq="M" instead of freq="ME" which gives an error.


marcopeix:
Hello! Thanks, but this was already done in #1274 . It has been merged already, and the current readme file shows the correct frequency to use.

Issue #712, opened 2023-08-01
[Loss] Poor performance with the NegativeBinomial DistributionLoss


Antoine-Schwartz:
### What happened + What you expected to happen

I suspect a bug around the binomial negative. Indeed, performance seems to be off compared with other available distributions, even when faced with positive count data on which it is supposed to be efficient.

Perhaps a conflict with the way the input data is scaled? I know that on Pytorch-Forecasting, they block the use of negative binomial when applying centered normalization: https://pytorch-forecasting.readthedocs.io/en/stable/_modules/pytorch_forecasting/metrics/distributions.html#NegativeBinomialDistributionLoss

I can't share the results on my data, but I've coded a quick example that illustrates the problem.

### Versions / Dependencies

neuralforecast==1.7.4
torch==2.3.1+cu121

### Reproduction script

```
import pandas as pd
import numpy as np 
import itertools

from neuralforecast import NeuralForecast
from neuralforecast.models import DeepAR, TFT, NHITS
from neuralforecast.losses.pytorch import DistributionLoss
from neuralforecast.losses.numpy import mae
from neuralforecast.utils import AirPassengersPanel

Y_df = AirPassengersPanel

nf = NeuralForecast(
    models=[
        eval(model)(
            h=12,
            input_size=48,
            max_steps=100,
            scaler_type="robust",
            loss=DistributionLoss(distr, level=[]),
            alias=f"{model}-{distr}",
            enable_model_summary=False,
            enable_checkpointing=False,
            enable_progress_bar=False,
            logger=False
        )
        for model, distr in itertools.product(
            ["DeepAR", "TFT", "NHITS"], ["Poisson", "Normal", "StudentT", "NegativeBinomial"]
        )
    ],
    freq="M"
)
cv_df = nf.cross_validation(Y_df, n_windows=5, step_size=12).reset_index();

def evaluate(df):
    eval_ = {}
    df = df.merge(Y_df[["unique_id", "ds", "y_[lag12]"]], how="left").rename(columns={"y_[lag12]": "seasonal_naive"})
    models = ["seasonal_naive"] + list(df.columns[df.columns.str.contains('median')])
    for model in models:
        eval_[model] = {}
        eval_[model][mae.__name__] = int(np.round(mae(df['y'].values, df[model].values), 0))
    eval_df = pd.DataFrame(eval_).rename_axis('metric')
    return eval_df

cv_df.groupby('cutoff').apply(lambda df: evaluate(df))
```
Output:
[...]

PR #1289, opened 2025-03-16
[Fix] Poor performance with the NegativeBinomial DistributionLoss

JQGoh:
An attempt to address https://github.com/Nixtla/neuralforecast/issues/712 , suspect that the wrong scaling parameters were chosen during the Negative Binomial distribution computation


Antoine-Schwartz:
Hello @JQGoh,

I'm glad someone is finally taking this subject seriously. Indeed, as I didn't have the time to dissect the code in detail to understand the problem, I had gone for the default `StudentT` in my codes.
Did you try to relaunch the example of my issue (it spins fast too) with your fix? I don't think I'll have time to test it this week, but I'll let you know as soon as I can.

However, my team has noticed that `v3.0.0` of neuralforecast has a negative impact on the overall performance of TFT (but maybe not NHITS?). We'll be opening an issue on this subject this week. If this is actually the case, it will make comparisons difficult: TFT `v2.0.1` with `StudentT` vs. TFT `v3.0.1` (including your fix) with `NegativeBinomial`.

Antoine-Schwartz:
I took a few minutes to test it out.
Your fix seems to work correctly from v3.0.0! Performance is similar to other distributions, if not better, as when using DeepAR.

marcopeix:
Also tested on my end, and I see an improvement with DeepAR, NHITS and TFT. Thanks a lot @JQGoh for the fix, and thank you @Antoine-Schwartz for testing!

PR #1292, opened 2025-03-19
add flash attention

LeonEthan:
- add flash attention (torch) in common modules.
- add param `atten` for model: TimeXer/iTransformer/VanillaTransformer, select from ['full', 'flash']
- add feat test on script in `tests/`


LeonEthan:
I don't know why, when I do a `git commit`, it seems that `nbdev` generates many modifications that I didn't make, even though most of those changes are just comments.

PR #1295, opened 2025-03-20
add flash-attention

LeonEthan:
- add flash attention (torch) in common modules.
- add param atten for model: TimeXer/iTransformer/VanillaTransformer, select from ['full', 'flash']


neuralforecast/common/_modules.py
```diff
@@ -378,33 +378,56 @@ def __init__(
...
         attention_dropout=0.1,
         output_attention=False,
+        atten="full",
     ):
         super(FullAttention, self).__init__()
         self.scale = scale
         self.mask_flag = mask_flag
         self.output_attention = output_attention
         self.dropout = nn.Dropout(attention_dropout)
+        self.atten = atten
```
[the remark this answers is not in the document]
LeonEthan: I tested installing the 3.0.0 version of the library using pip, trained and saved the model. Then I installed it using the branch code and tested loading the model for prediction and training. So far, everything seems to work. I think the `atten` parameter is not explicitly passed, which won't cause this problem. Moreover, the "full" process in the original code remains unchanged. Just to be safe, I still added a `_load_from_state_dict` function and submitted a commit: [8e93aae].

marcopeix:
Hello! Thanks a lot for the contribution! 

We try to stick to the implementation from the papers, and we haven't done flash attention as it is not used in the original architectures. Usually, for this kind of change, we want to see some added benefit to diverge from the original architecture.

Could you run a small benchmark to compare between using full attention vs flash attention? Doesn't have to be big, but if you can report results on the ETT datasets (ETTm1, ETTm2, ETTh1, ETTh2), that would really help assess if adding flash attention is worth it. Maybe using TimeXer and iTransformer, they should train pretty fast. You can take inspiration from [here](https://github.com/marcopeix/time-series-analysis/blob/master/itransformer_blog.ipynb), where I pretty much setup the experiment, just need to modify the models. 

Thanks!

elephaint:
> Hello! Thanks a lot for the contribution!
> 
> We try to stick to the implementation from the papers, and we haven't done flash attention as it is not used in the original architectures. Usually, for this kind of change, we want to see some added benefit to diverge from the original architecture.
> 
> Could you run a small benchmark to compare between using full attention vs flash attention? Doesn't have to be big, but if you can report results on the ETT datasets (ETTm1, ETTm2, ETTh1, ETTh2), that would really help assess if adding flash attention is worth it. Maybe using TimeXer and iTransformer, they should train pretty fast. You can take inspiration from [here](https://github.com/marcopeix/time-series-analysis/blob/master/itransformer_blog.ipynb), where I pretty much setup the experiment, just need to modify the models.
> 
> Thanks!

I don't think FlashAttention is different than normal attention, it's just more memory efficient. See the paper [here](https://arxiv.org/abs/2205.14135).

@JQGoh @LeonEthan Why are we not simply using Pytorch's `TransformerEncoderLayer()`, replacing all these custom modules? [That one also will use flash attention by default if eligible](https://github.com/pytorch/pytorch/blob/1eba9b3aa3c43f86f4a2c807ac8e12c4a7767340/torch/nn/modules/transformer.py#L630).

I also don't think we should be adding an additional parameter for this - if it can do flash attention, it should always do it, as it's the same as full attention but just faster. Unless you're proposing we use block-sparse flash attention, but that's not in the code that's been proposed in this PR, if I'm not mistaken. 

I'd do:
- Replace these custom transformer modules by a call to PyTorch's `TransformerEncoderLayer()`, removing a lot of code from our Transformer-based models.
- Speed test the result - the only reason I can think of not to switch to PyTorch's built-in Transformer layers is that they tend to have more boilerplate / overhead, so performance (speed) may be a concern.

LeonEthan:
> > Hello! Thanks a lot for the contribution!
> > We try to stick to the implementation from the papers, and we haven't done flash attention as it is not used in the original architectures. Usually, for this kind of change, we want to see some added benefit to diverge from the original architecture.
> > Could you run a small benchmark to compare between using full attention vs flash attention? Doesn't have to be big, but if you can report results on the ETT datasets (ETTm1, ETTm2, ETTh1, ETTh2), that would really help assess if adding flash attention is worth it. Maybe using TimeXer and iTransformer, they should train pretty fast. You can take inspiration from [here](https://github.com/marcopeix/time-series-analysis/blob/master/itransformer_blog.ipynb), where I pretty much setup the experiment, just need to modify the models.
> > Thanks!
> 
> @marcopeix I don't think FlashAttention is different than normal attention, it's just more memory access efficient. See the paper [here](https://arxiv.org/abs/2205.14135).
> 
> @JQGoh @LeonEthan Why are we not simply using Pytorch's `TransformerEncoderLayer()`, replacing all these custom modules? [That one also will use flash attention by default if eligible](https://github.com/pytorch/pytorch/blob/1eba9b3aa3c43f86f4a2c807ac8e12c4a7767340/torch/nn/modules/transformer.py#L682).
> 
> I also don't think we should be adding an additional parameter for this - if it can do flash attention, it should always do it, as it's the same as full attention but just faster. Unless you're proposing we use block-sparse flash attention, but that's not in the code that's been proposed in this PR, if I'm not mistaken.
> 
> I'd do:
> 
> * Replace these custom transformer modules by a call to PyTorch's `TransformerEncoderLayer()`, removing a lot of code from our Transformer-based models.
> * Speed test the result - the only reason I can think of not to switch to PyTorch's built-in Transformer layers is that they tend to have more boilerplate / overhead, so performance (speed) may be a concern. This is automatically done in our CI tests so you don't have to explicitly do this (we test both speed and accuracy of e.g. VanillaTransformer, so if you include it for that model we can see the performance impact in our CI tests).

The current `TransEncoderLayer` in the code offers greater flexibility, allowing it to be reused in many models. For example, the SOFTS model also only uses `TransEncoderLayer`, the attention parameters it passes in are a `STAD` module, not a conventional attention module. PyTorch's `TransformerEncoderLayer` might not be able to replace it in this case. 
BTW, I also agree that adding parameters is somewhat inelegant. Perhaps there is a better way.

elephaint:
> > > Hello! Thanks a lot for the contribution!
> > > We try to stick to the implementation from the papers, and we haven't done flash attention as it is not used in the original architectures. Usually, for this kind of change, we want to see some added benefit to diverge from the original architecture.
> > > Could you run a small benchmark to compare between using full attention vs flash attention? Doesn't have to be big, but if you can report results on the ETT datasets (ETTm1, ETTm2, ETTh1, ETTh2), that would really help assess if adding flash attention is worth it. Maybe using TimeXer and iTransformer, they should train pretty fast. You can take inspiration from [here](https://github.com/marcopeix/time-series-analysis/blob/master/itransformer_blog.ipynb), where I pretty much setup the experiment, just need to modify the models.
> > > Thanks!
> > 
> > 
> > @marcopeix I don't think FlashAttention is different than normal attention, it's just more memory access efficient. See the paper [here](https://arxiv.org/abs/2205.14135).
> > @JQGoh @LeonEthan Why are we not simply using Pytorch's `TransformerEncoderLayer()`, replacing all these custom modules? [That one also will use flash attention by default if eligible](https://github.com/pytorch/pytorch/blob/1eba9b3aa3c43f86f4a2c807ac8e12c4a7767340/torch/nn/modules/transformer.py#L682).
> > I also don't think we should be adding an additional parameter for this - if it can do flash attention, it should always do it, as it's the same as full attention but just faster. Unless you're proposing we use block-sparse flash attention, but that's not in the code that's been proposed in this PR, if I'm not mistaken.
> > I'd do:
> > 
> > * Replace these custom transformer modules by a call to PyTorch's `TransformerEncoderLayer()`, removing a lot of code from our Transformer-based models.
> > * Speed test the result - the only reason I can think of not to switch to PyTorch's built-in Transformer layers is that they tend to have more boilerplate / overhead, so performance (speed) may be a concern. This is automatically done in our CI tests so you don't have to explicitly do this (we test both speed and accuracy of e.g. VanillaTransformer, so if you include it for that model we can see the performance impact in our CI tests).
> 
> The current `TransEncoderLayer` in the code offers greater flexibility, allowing it to be reused in many models. For example, the SOFTS model also only uses `TransEncoderLayer`, the attention parameters it passes in are a `STAD` module, not a conventional attention module. PyTorch's `TransformerEncoderLayer` might not be able to replace it in this case. BTW, I also agree that adding parameters is somewhat inelegant. Perhaps there is a better way.

Ok, happy to add it then to modules but not with the parameter imho - if a model requests the attention scores `A`, it should go to full attention, otherwise flash should be the default if it's the same.

Doesn't this also apply to PatchTST, TFT and *Former models? I think we should change it there too. 

Do you have a micro experiment demonstrating output equivalence to the dense method?

LeonEthan:
@marcopeix @elephaint 

Hi, I've updated the PR with the requested changes to remove the parameter and set flash attention as default, also added it for PatchTST. Ready for review when you have a chance. Thanks!

BTW，I have considered the TFT and xxformer models, and I think they don't seem suitable for modification. For xxformer, it's because they usually use a different SDPA computation method, and for TFT, it's because the attention weights are important to TFT, but flash-attention is not supported.

nbs/models.patchtst.ipynb
```diff
@@ -570,33 +570,50 @@
...
     "            scores : [bs x n_heads x q_len x seq_len]\n",
     "        '''\n",
+    "        if not self.res_attention:\n",
+    "            # Use torch's built-in flash attention for efficient computation\n",
+    "            # Note: This will not return attention weights/scores\n",
+    "            # The shapes of q, k, v must be: [batch, n_heads, seq_len, head_dim]\n",
+    "            # Reshape q, k, v into [batch*n_heads, seq_len, head_dim] as required by torch.nn.functional.scaled_dot_product_attention\n",
+    "            bs, n_heads, seq_len, head_dim = q.shape\n",
+    "            q_ = q.reshape(bs * n_heads, seq_len, head_dim)\n",
+    "            k_ = k.reshape(bs * n_heads, seq_len, head_dim)\n",
```
elephaint, line 580: Check the input shapes, this should be:

`k_ = k.permute(0, 1, 3, 2).reshape(bs * n_heads, seq_len, head_dim)`

nbs/models.patchtst.ipynb
```diff
@@ -570,33 +570,50 @@
...
+    "            v_ = v.reshape(bs * n_heads, seq_len, head_dim)\n",
+    "            # If attn_mask exists, convert it to the appropriate format for flash attention (e.g. [batch*n_heads, seq_len, seq_len])\n",
+    "            if attn_mask is not None:\n",
+    "                attn_mask = attn_mask.repeat(bs * n_heads, 1, 1)\n",
+    "            output = torch.nn.functional.scaled_dot_product_attention(\n",
```
elephaint, line 585: We already import `torch.nn.functional as F`, so let's make use of it to keep code readable:

`output = F.scaled_dot_product_attention(`

nbs/common.modules.ipynb
```diff
@@ -478,23 +478,34 @@
...
-    "            if attn_mask is None:\n",
-    "                attn_mask = TriangularCausalMask(B, L, device=queries.device)\n",
+    "        \n",
+    "        if not self.output_attention:   # flash attention not supported\n",
+    "            q = queries.permute(0, 2, 1, 3)  # [B, H, L, E]\n",
+    "            k = keys.permute(0, 2, 1, 3)\n",
+    "            v = values.permute(0, 2, 1, 3)\n",
+    "            \n",
+    "            scale = self.scale or 1.0 / math.sqrt(E)\n",
+    "            attn_output = torch.nn.functional.scaled_dot_product_attention(\n",
```
elephaint, line 488: We already import `torch.nn.functional as F`, so let's make use of it to keep code readable:

`output = F.scaled_dot_product_attention(`

elephaint (changes requested):
Thanks! Few small comments. 

Great that you added the test - to me it seems the differences are indeed negligible, so I'd be confident implementing it like this. What do you think?

nbs/models.patchtst.ipynb
LeonEthan: Done, fixed in the latest commit.

nbs/models.patchtst.ipynb
LeonEthan: Done, fixed in the latest commit.

LeonEthan: Done, fixed in the latest commit.

LeonEthan:
> Contributor

@elephaint  Hi, I've made the requested changes based on your review and pushed a new commit to the branch.



> Thanks! Few small comments.
> 
> Great that you added the test - to me it seems the differences are indeed negligible, so I'd be confident implementing it like this. What do you think about the test results?

Regarding the test results, I believe the observed differences in error are even smaller than what might be attributed to random variation or noise. Such minor discrepancies are generally within an acceptable range. Based on my personal experience, time series scenarios often involve inherent noise, and blindly pursuing extremely small changes in metrics can easily lead to overfitting. 
As for training time, I also conducted more tests on my company's datasets. On V100/A10 GPUs, Flash Attention achieved a 10% to 20% reduction in training time, which would be significantly advantageous in large-scale data scenarios.

elephaint (approved):
LGTM, thanks for your work @LeonEthan! Feel free to merge.

LeonEthan:
Hi @elephaint @marcopeix 

I don’t have write access to merge this PR. Could you please help merge it when you have a chance? Thanks!

PR #1306, opened 2025-03-31
FIX: Add exclude_insample_y param to TimeXer for model loading

marcopeix:
TimeXer cannot be loaded because it didn't handle the `exclude_insample_y` parameter correctly. I added it to the `init` of the model and loading works now.


neuralforecast/_modidx.py
```diff
@@ -517,60 +517,7 @@
                                                                                                          'neuralforecast/losses/pytorch.py'),
                                                'neuralforecast.losses.pytorch.weighted_average': ( 'losses.pytorch.html#weighted_average',
                                                                                                    'neuralforecast/losses/pytorch.py')},
-            'neuralforecast.models.autoformer': { 'neuralforecast.models.autoformer.AutoCorrelation': ( 'models.autoformer.html#autocorrelation',
```
elephaint, line 520: This change is wrong I think - this shouldn't be deleted, which is the reason the nbdev version was previously pinned to 2.3.25. 

We need to overcome this issue before upgrading the nbdev version.

elephaint (changes requested):
I think we need to find a fix to nbdev not processing auto* files before upgrading the version.

PR #1307, opened 2025-04-01
[FEAT] HuberIQLoss

elephaint:
Huberized version of `IQLoss`.

Basically a copy of `IQLoss`, only it inherits from `HuberQLoss` rather than from `QuantileLoss`.


marcopeix (approved):
LGTM

PR #1309, opened 2025-04-16
Distributional predictions in `predict_insample()`

janovergoor:
* adds option to keep distributional predictions in `predict_insample()` with `keep_distribution=True` (default is `False`)
* fixes an issue in `predict_insample()` for variations on `DistributionLoss`, where the default behavior was to return the sample mean rather than the median (which is what the documentation says it does).
* adds tests for this new use of  `predict_insample()`
* fixes a few typos in comments/documentation


marcopeix:
Hi @janovergoor, thanks for the PR, looks like a great feature!

Did you follow our [contribution guidelines](https://github.com/Nixtla/neuralforecast/blob/main/CONTRIBUTING.md)? I see you modified `core.ipynb`, but usually if you run `nbdev_export`, we should also see the Python file has been modified. Just make sure to follow the guidelines and let me know if you need some help!

janovergoor:
Thanks @marcopeix  - I hadn't read the instructions but have now and updated the `.py` file with `nbdev_export` also.

marcopeix:
@janovergoor Very good job on this PR! Everything looks great, I would just like to add an example in the tutorial. 

In the notebook `nbs/docs/capabilities/05_predictInsample.ipynb`, can you add a new section "6. Insample predictions with prediction intervals". 

Then, add a cell to import distribution losses:
```
```

Then, add a cell with this code:
```python
horizon = 12

# Try different hyperparmeters to improve accuracy.
models = [NHITS(h=horizon,
                input_size=2 * horizon,
                loss=DistributionLoss(distribution="Poisson", num_samples=50),
                max_steps=100,
                n_freq_downsample=[2, 1, 1],
                mlp_units = 3 * [[1024, 1024]],
               )
          ]
nf = NeuralForecast(models=models, freq='ME')
nf.fit(df=Y_df, val_size=horizon)
Y_hat_insample = nf.predict_insample(
    step_size=horizon, 
    keep_distribution=True # Get the insample prediction intervals
)
plot_series(forecasts_df=Y_hat_insample.drop(columns=['cutoff', 'NHITS-median']), level=[80, 90])
```
After, make sure to `nbdev_clean --fname nbs/docs/capabilities/05_predictInsample.ipynb` before pushing the code, so the tests pass.

Once this is done and all tests pass, I'll merge it. Again, great work on this PR. Let me know if you have any questions!

janovergoor:
Thanks! Documentation added

nbs/core.ipynb
```diff
@@ -1360,8 +1362,8 @@
     "                model.set_test_size(test_size=trimmed_dataset.max_size)\n",
     "                # Generate predictions\n",
     "                model_fcsts = model.predict(trimmed_dataset, step_size=step_size)\n",
-    "                # Handle distributional forecasts; take only median\n",
-    "                if len(model_fcsts.shape) > 1 and model_fcsts.shape[1] == 3:\n",
+    "                if not keep_distribution and len(model_fcsts.shape) > 1 and model_fcsts.shape[1] == 3:\n",
```
elephaint, line 1365: We shouldn't do this like this, because this will fail on multivariate models.

Please test this functionality for:
- Every possible loss function
- Every possible architecture (LSTM, TSMixer, NHITS)

I need to think about how to incorporate it properly.

elephaint (changes requested):
Nice! I think we should do the if statement different though - we should test this functionality for all types of architectures and all types of probabilistic losses:

- TSMixer, LSTM and NHITS
- Test against all probabilistic losses.

nbs/core.ipynb
janovergoor: would it work to use `not keep distribution and hasattr(model.loss, 'quantiles')` for the `if` condition instead?
that seems to caputure all the loss functions that return more than one value for `predict()` that I could find.

janovergoor: see how you like it now? I implemented the above and added tests.

One thing to note is that because for variations on `DistributionLoss` (also `PMM`, `GMM`, `NBMM`), `predict()` returns the sample mean as well as the median:
* the default behavior of `predict_insample()` used to return the sample mean (as it's the first column) - now it returns the median to be consistent with the other losses and the documented behavior of returning the median.
* currently, `keep_distribution=True` will return all of the sample mean, median, and levels for variations of `DistributionLoss`

elephaint:
@janovergoor Thanks for the changes, I'll have a look this week!

marcopeix:
> ping on this @elephaint 🙏

@janovergoor he's on vacation, I can see if I have time to take a look at it this week.

nbs/core.ipynb
elephaint: Sorry for the delay, I'm looking at it!

elephaint:
@janovergoor Sorry for the delay, I'm looking at it.

elephaint:
@janovergoor I was playing with it, and found that some loss functions still errored (e.g. `IQLoss`). That made me realize that it's perhaps better to have the API of `predict_insample` be consistent with `predict`, meaning:

- The default behaviour of `predict_insample` should be to return the same as `.predict` (so with levels and median)
- Add a `level` and `quantiles` argument to `predict_insample`. This allows to return any quantile / level, akin to how it can be done in `.predict`

There's a few issues with this idea though, so I need some more time with it, and curious for your thoughts:
- It doesn't work yet with multivariate models, because we loop over each dataset. It makes me wonder how predict_insample currently works for multivariate models...
- I need to change the test remove the argument

It's just an idea, I don't know. E.g. it still doesn't work with conformal though, so maybe it's just not a good idea....

elephaint:
> @janovergoor I was playing with it, and found that some loss functions still errored (e.g. `IQLoss`). That made me realize that it's perhaps better to have the API of `predict_insample` be consistent with `predict`, meaning:
> 
>     * The default behaviour of `predict_insample` should be to return the same as `.predict` (so with levels and median)
> 
>     * Add a `level` and `quantiles` argument to `predict_insample`. This allows to return any quantile / level, akin to how it can be done in `.predict`
> 
> 
> There's a few issues with this idea though, so I need some more time with it, and curious for your thoughts:
> 
>     * It doesn't work yet with multivariate models, because we loop over each dataset. It makes me wonder how predict_insample currently works for multivariate models...
> 
>     * I need to change the test remove the argument
> 
> 
> It's just an idea, I don't know. E.g. it still doesn't work with conformal though, so maybe it's just not a good idea....

@janovergoor 
So I found out that insample predictions for multivariate models are basically nonsense at the moment 😬 So, for now I've simply excluded them, raising an informative error.

I think the changes now cover your use case and raise an error where appropriate.

elephaint (approved):
@janovergoor I think it's good to go like this; the behavior is that you can get the levels akin to how you'd get the levels in `.predict`, i.e. by specifying the required level or quantile. 

This also makes it generalize over any level or any quantile requested post-training, whereas with the `keep_distribution` qualifier this wouldn't have been possible easily.

janovergoor:
thanks for pulling this through!

PR #1317, opened 2025-04-23
FEAT: Add the option of selecting an optimal starting learning rate using `lr_find`

marcopeix:
The tuner in Pytorch Lightning has an `lr_find` method that selects an optimal starting learning rate. This PR adds the `enable_lr_finder` parameter to all models to find the optimal learning rate.

Thanks to @carusyte for PR #1263 , I used it as a starting point for this PR.


environment-cpu.yml
```diff
@@ -15,7 +15,7 @@ dependencies:
   - optuna
   - pandas>=1.3.5
   - pyarrow
-  - pytorch>=2.0.0
+  - pytorch<=2.6.0
```
elephaint, line 18: This should have as minimum 2.0.0, I think the syntax is like this:

`pytorch>=2.0.0,<=2.6.0` or `"pytorch>=2.0.0,<=2.6.0"`

environment-cuda.yml
```diff
@@ -15,7 +15,7 @@ dependencies:
   - optuna
   - pandas>=1.3.5
   - pyarrow
-  - pytorch>=2.0.0
+  - pytorch<=2.6.0
```
elephaint, line 18: See above, minimum should still be 2.0.0

elephaint (changes requested):
This entire PR isn's necessary imho, we can just do:

```
from pytorch_lightning.callbacks import LearningRateFinder
models = [NHITS(h=12, input_size=24, max_steps=10, learning_rate=1e-3, callbacks=[LearningRateFinder()])]
```

so I don't see the need for this PR? 

Every model already has **trainer_kwargs as argument, so we can just pass pl trainer callbacks and arguments to every model already.

PR #1318, opened 2025-04-24
FIX: Set 2.0.0<=pytorch<=2.6.0 to avoid conflicts with networkx with Python 3.9

marcopeix:
Installing PyTorch 2.7 creates a conflict with networkx when using Python 3.9.

Fixing the version to be between 2 and 2.6 solves the problem.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,7 +36,7 @@ jobs:
       - name: Install pip requirements
         run: >
           pip install uv &&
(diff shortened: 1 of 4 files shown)


elephaint (approved):
Nice, LGTM!

candalfigomoro:
@marcopeix @elephaint Please remember to also update the conda recipe accordingly https://github.com/conda-forge/neuralforecast-feedstock

PR #1319, opened 2025-04-24
FIX: Allow static vars only with NBEATSx and exogenous block

marcopeix:
Currently, using static variables only and specifying an 'exogenous' stack causes an error with NBEATSx.

```python
from neuralforecast.utils import AirPassengersDF, AirPassengersPanel, AirPassengersStatic
from neuralforecast.models import NBEATSx

Y_train_df = AirPassengersPanel[AirPassengersPanel.ds<AirPassengersPanel['ds'].values[-12]]
Y_test_df = AirPassengersPanel[AirPassengersPanel.ds>=AirPassengersPanel['ds'].values[-12]].reset_index(drop=True)

model = NBEATSx(h=12, input_size=24,
                scaler_type='robust',
                dropout_prob_theta=0.5,
                stat_exog_list=['airline1'],
                stack_types = ["identity", "trend", "seasonality", "exogenous"],
                n_blocks = [1,1,1,1],
                max_steps=200,
                val_check_steps=10,
                early_stop_patience_steps=2)

nf = NeuralForecast(
    models=[model],
    freq='ME'
)
nf.fit(df=Y_train_df, static_df=AirPassengersStatic, val_size=12)
```
There are two possible solutions:
1. Remove the 'exogenous' stack when using static variables only. However, I'm not sure that the variables will be actually used in training.
2. Reshape static variables when there are not futr_exog. This is what this PR implements.

### DIFF
--- a/nbs/models.nbeatsx.ipynb
+++ b/nbs/models.nbeatsx.ipynb
@@ -3,6 +3,16 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "id": "1e3afcc3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%set_env PYTORCH_ENABLE_MPS_FALLBACK=1"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
    "id": "524620c1",
    "metadata": {},
    "outputs": [],
@@ -282,7 +292,9 @@
     "        \"\"\" \"\"\"\n",
     "        super().__init__()\n",
     "\n",
+    "        self.h = h\n",
     "        self.dropout_prob = dropout_prob\n",
+    "        self.input_size = input_size\n",
     "        self.futr_input_size = futr_input_size\n",
     "        self.hist_input_size = hist_input_size\n",
     "        self.stat_input_size = stat_input_size\n",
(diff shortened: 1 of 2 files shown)


nbs/models.nbeatsx.ipynb
```diff
@@ -392,7 +404,7 @@
     "    `exclude_insample_y`: bool=False, the model skips the autoregressive features y[t-input_size:t] if True.<br>\n",
     "    `n_harmonics`: int, Number of harmonic oscillations in the SeasonalityBasis [cos(i * t/n_harmonics), sin(i * t/n_harmonics)]. Note that it will only be used if 'seasonality' is in `stack_types`.<br>\n",
     "    `n_polynomials`: int, Number of polynomial terms for TrendBasis [1,t,...,t^n_poly]. Note that it will only be used if 'trend' is in `stack_types`.<br>\n",
-    "    `stack_types`: List[str], List of stack types. Subset from ['seasonality', 'trend', 'identity'].<br>\n",
+    "    `stack_types`: List[str], List of stack types. Subset from ['seasonality', 'trend', 'identity', 'exogenous'].<br>\n",
```
elephaint, line 407: What does this do? Because it seems we're not filtering on `exogenous` anywhere?

elephaint (changes requested):

nbs/models.nbeatsx.ipynb
elephaint: Never mind, I just read the related issue, I understand that it's missing from the documentation and it should be included here.

elephaint (approved):

PR #1325, opened 2025-05-13
[FIX] Create windows once

elephaint:
Currently, we create windows on every iteration in a prediction- and validation loop. However, we can just create the windows once before entering the batch loop, and then sample from them in the loop. In terms of memory this should have zero performance impact, we save a bit of unnecessary control flow / compute. 

Edit: from our CircleCI tests, performance increases somewhat (i.e., our CI tests run faster by about 5%). Expect more performance improvement if there's more validation/prediction in the code.


marcopeix (approved):
Nice!

PR #1335, opened 2025-05-22
Optimizations in tsdataset: reduce allocations for large datasets

tylernisonoff:
When using neuralforecast with a large dataset and large batches, I noticed the GPU was being heavily underutilized. 

These two optimizations resulted in a 10x speed-up for my training runs. 

1. avoid allocations in `_collate_fn` if `len(batch) == 1`

2. avoid extra allocation of temporal if padding is not needed in `__getitem__`

The length of my dataset is `152_809_920`. My NHITS model had specified batch_size of 32, but for whatever reason the batch in `_collate_fn` was always of length 1. So a bit unclear if there is some other bug putting me in a state where the first optimization applies. Regardless, these optimizations seem relatively harmless to code complexity, while reducing allocations.


tylernisonoff:
```
Cells with imports and code..............................................Passed
nbdev_clean..............................................................Passed
nbdev_export.............................................................Failed
- hook id: nbdev_export
- files were modified by this hook
```

when I run `nbdev_clean --clear_all`, it modifies a ton of notebooks, none of which I changed, so I suspect this is causing the failure? 

```
nbdev_clean
nbdev_export
```
is clean for me locally

PR #1340, opened 2025-06-12
FIX: Add logic to load custom models when using ReduceLROnPlateau

marcopeix:
Currently, we have a tutorial that shows how to use `ReduceLROnPlateau` which requires creating a custom model class:
```python
class CustomNHITS(NHITS):
        optimizer = torch.optim.Adadelta(params=self.parameters(), rho=0.75)
        scheduler=torch.optim.lr_scheduler.ReduceLROnPlateau(
            optimizer=optimizer, mode='min',factor=0.5, patience=2,
        )
        scheduler_config = {
            'scheduler': scheduler,
            'interval': 'step',
            'frequency': 1,
            'monitor': 'train_loss',
            'strict': True,
            'name': None,
        }
        return {'optimizer': optimizer, 'lr_scheduler': scheduler_config}
```
However, this model can be saved and loaded, because `CustomNHITS` is not in `MODEL_FILENAME_DICT`. 

This PR fixes that issue by looking for the best match in `MODEL_FILENAME_DICT`. 

For example, `CustomNHITS` would match the `nhits` key, and so we load the NHITS model.


elephaint:
Pfff.... this is a hard one. I think we can do it simpler. We just need to save the right base class in `alias_to_model.pkl`, right?

So, the only change I think we need is this - change [this line](https://github.com/Nixtla/neuralforecast/blob/0f16b3a8c93ef67be2fed28c5257e5098cfb4815/neuralforecast/core.py#L1467) to this:

```
if model.__class__.__name__.lower() in MODEL_FILENAME_DICT:
    model_class_name = model.__class__.__name__.lower()
elif model.__class__.__base__.__name__.lower() in MODEL_FILENAME_DICT:
    model_class_name = model.__class__.__base__.__name__.lower()
else:
    raise ValueError(
        f"Model {model.__class__.__name__} is not supported for saving."
    ))
```

With the above change the following example succeeds on my machine (and fails without):
```

import pandas as pd
import torch

Y_train_df = AirPassengersPanel[AirPassengersPanel.ds<AirPassengersPanel['ds'].values[-12]].reset_index(drop=True) # 132 train
Y_test_df = AirPassengersPanel[AirPassengersPanel.ds>=AirPassengersPanel['ds'].values[-12]].reset_index(drop=True) # 12 test

class CustomNHITS(NHITS):
        optimizer = torch.optim.Adadelta(params=self.parameters(), rho=0.75)
        scheduler=torch.optim.lr_scheduler.ReduceLROnPlateau(
            optimizer=optimizer, mode='min',factor=0.5, patience=2,
        )
        scheduler_config = {
            'scheduler': scheduler,
            'interval': 'step',
            'frequency': 1,
            'monitor': 'train_loss',
            'strict': True,
            'name': None,
        }
        return {'optimizer': optimizer, 'lr_scheduler': scheduler_config}

model = CustomNHITS(
    input_size=24,
    h =5,
    max_steps=2,
)

model2 = NHITS(
    input_size=24,
    h =5,
    max_steps=2,
)

nf = NeuralForecast(models=[model, model2], freq='H')
nf.fit(df=Y_train_df)


nf.save("tmp/checkpoints/", overwrite=True, save_dataset=False)

nf = NeuralForecast.load("tmp/checkpoints/")
```

elephaint (changes requested):
See my other comment, I think we can do this simpler in 4 lines of code

elephaint (approved):
Nice! LGTM!

PR #1344, opened 2025-06-19
FEAT: Training window filtering

marcopeix:
Currently, we create the maximum number of training windows, meaning that we might have windows with only 1 available insample data point and 1 available outsample data point. 

These are technically low quality windows.

This PR adds the parameter `available_sample_fractions` to control how many available insample and outsample data points should be available as a fraction of input size and horizon for insample and outsample respectively.


marcopeix:
It's basically a rework of #1059 . Many thanks to @jasminerienecker for the initial idea!

nbs/common.base_model.ipynb
```diff
@@ -358,6 +370,16 @@
     "        else:\n",
     "            self.inference_windows_batch_size = inference_windows_batch_size\n",
     "\n",
+    "        # Filtering training windows by available sample fractions\n",
+    "        if isinstance(available_sample_fractions, (int, float)):\n",
```
elephaint: It can't be an int, can it? The definition states it needs to be a float.

nbs/common.base_model.ipynb
```diff
@@ -358,6 +370,16 @@
...
+    "            self.min_insample_fraction = float(available_sample_fractions)\n",
+    "            self.min_outsample_fraction = float(available_sample_fractions)\n",
+    "        elif isinstance(available_sample_fractions, (list, tuple)) and len(available_sample_fractions) == 2:\n",
+    "            self.min_insample_fraction = float(available_sample_fractions[0])\n",
+    "            self.min_outsample_fraction = float(available_sample_fractions[1])\n",
+    "        else:\n",
+    "            raise ValueError(\"available_sample_fractions must be a float or a list/tuple of two floats\")\n",
```
elephaint: ~~This needs protection against `1.0`, `[1.0, 1.0]`, `[1.0, 0.0]`, `[0.0, 1.0]`~~

~~This needs protection against what the implication is of the number of samples multiplied by the fraction. E.g. 2 * 0.99 might still be too few samples~~

Never mind, you're protecting by setting a minimum later on, so this it moot.

It does need protection against values >1.0, I don't think these are covered in the code below

nbs/common.base_model.ipynb
```diff
@@ -674,16 +696,20 @@
...
     "                windows = windows.unsqueeze(-1)\n",
     "\n",
-    "            # Sample and Available conditions\n",
-    "            available_idx = temporal_cols.get_loc('available_mask')           \n",
-    "            available_condition = windows[:, :self.input_size, available_idx]\n",
-    "            available_condition = torch.sum(available_condition, axis=(1, -1)) # Sum over time & series dimension\n",
-    "            final_condition = (available_condition > 0)\n",
-    "            \n",
+    "            # Calculate minimum required available points based on fractions\n",
+    "            min_insample_points = max(1, int(self.input_size * self.min_insample_fraction))\n",
```
elephaint: What happens when `min_insample_fraction>1.0`?

nbs/common.base_model.ipynb
```diff
@@ -27,7 +27,18 @@
    "execution_count": null,
    "id": "1c7c2ba5-19ee-421e-9252-7224b03f5201",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
elephaint: Clear nb

nbs/common.base_model.ipynb
```diff
@@ -1426,6 +1452,142 @@
...
+    "    \n",
+    "    try:\n",
+    "        # Should raise error for invalid type\n",
+    "        process_sample_fractions(\"invalid\")\n",
+    "        assert False, \"Should have raised ValueError for invalid type\"\n",
+    "    except ValueError:\n",
+    "        pass\n",
+    "    \n",
+    "    # Test valid inputs\n",
+    "    try:\n",
```
elephaint: Should include test with values larger than 1.0

nbs/common.base_model.ipynb
```diff
@@ -1426,6 +1452,142 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "b10c68fe",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| hide,\n",
+    "# Test window filtering\n",
+    "\n",
+    "def test_sample_fractions_parameter():\n",
```
elephaint: I think this is a good test but can you also test it fully in the sense of supplying an actual dataframe to an initialized NF model with the parameter varied?

Even though this tests the logic it doesn't test the logic for any of our models. 

Test should include NHITS, LSTM(recurrent=True), TSMixer.

nbs/models.autoformer.ipynb
```diff
@@ -458,6 +458,7 @@
     "    `windows_batch_size`: int=1024, number of windows to sample in each training batch, default uses all.<br>\n",
     "    `inference_windows_batch_size`: int=1024, number of windows to sample in each inference batch.<br>\n",
     "    `start_padding_enabled`: bool=False, if True, the model will pad the time series with zeros at the beginning, by input size.<br>\n",
+    "    `available_sample_fractions`: Union[float, List[float]]=0.0, minimum fraction of valid data points required for training windows. Single float applies to both insample and outsample; list of two floats specifies [insample_fraction, outsample_fraction]. Default 0.0 allows windows with only 1 valid data point (current behavior).<br>\n",
```
elephaint: I find the terminology a bit confusing. I'm more in favour of calling this a `data_availability_threshold` (see the original PR).

Suggestion:

`data_availability_threshold`:  Union[float, List[float]]=0.0, drop training windows where the percentage of available data points is less than this threshold. A single float drops the same percentage for both in- and outsample data points in the window. A list of two floats specifies [insample_data_availability_threshold, outsample_data_availability_threshold], allowing a different threshold for in- and outsample data points in the training window. Default 0.0 allows windows with only 1 valid data point (current behavior).

Note that we only apply it to training data at the moment, so maybe it should be called `training_data_availability_threshold`.

marcopeix: Agreed

nbs/common.base_model.ipynb
```diff
@@ -358,6 +359,28 @@
     "        else:\n",
     "            self.inference_windows_batch_size = inference_windows_batch_size\n",
     "\n",
+    "        # Filtering training windows by available sample fractions\n",
```
elephaint, line 362: 👍

nbs/common.base_model.ipynb
```diff
@@ -674,16 +697,20 @@
...
     "\n",
-    "            # Sample and Available conditions\n",
-    "            available_idx = temporal_cols.get_loc('available_mask')           \n",
-    "            available_condition = windows[:, :self.input_size, available_idx]\n",
-    "            available_condition = torch.sum(available_condition, axis=(1, -1)) # Sum over time & series dimension\n",
-    "            final_condition = (available_condition > 0)\n",
-    "            \n",
+    "            # Calculate minimum required available points based on fractions\n",
+    "            min_insample_points = max(1, int(self.input_size * self.min_insample_fraction))\n",
+    "            min_outsample_points = max(1, int(self.h * self.min_outsample_fraction))\n",
```
elephaint: I think should be `max(1, int(self.h * self.min_outsample_fraction * self.n_series)` because we sum over the time & series dimension to arrive at `available_condition`

nbs/common.base_model.ipynb
elephaint: The test still doesn't actually test the implemented behavior. 

A test should test the implemented behavior. This doesn't do that - it only validates some logic but it doesn't actually test what you implemented.

I'd delete this test entirely (because it has no purpose), and write a `test_sample_fraction.py` file, in which you create a pytest that tests the behavior. This will also fit into the new way of working when nbdev is removed. In addition, it allows you to test the full functionality for NHITS / LSTM / TSMixer type models.

elephaint (changes requested):
Thanks, good work! 2 things:

- It needs a proper test (the current one has no value)
- Minor adjustment in the implementation is required for multivariate models I think (which becomes apparent in a test)

elephaint (approved):
Great work! Good for me if you agree with the final changes I made (cosmetic and I changed the tests + added explanation to the tests)

nbs/core.ipynb
```diff
@@ -3676,6 +3676,319 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fe79d19f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| hide\n",
+    "# We are given m timeseries with n observations each, and we want to create a univariate forecast model with horizon h, input_size s, and step_size k. We will create windows of length h + s to train our model, and we will slide the window with step_size k.\n",
```
elephaint, line 3794: @marcopeix Check/verify

nbs/core.ipynb
```diff
@@ -3676,6 +3676,319 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "897e02e4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| hide\n",
+    "# We are given m timeseries with n observations each, and we want to create a multivariate forecast model with horizon h, input_size s, and step_size k. We will create windows of length h + s to train our model, and we will slide the window with step_size k.\n",
```
elephaint, line 3932: @marcopeix Check/verify

elephaint (commented):
@marcopeix good to go once you verify/check my explanations in the test

nbs/core.ipynb
marcopeix: Amazing explanation!

nbs/core.ipynb
marcopeix: Same as above

elephaint (approved):

Antoine-Schwartz:
Sampling quality is a really interesting subject. Thank you for taking the time to add this first option @marcopeix !

By the way, it would be great to have a dedicated documentation section. 
Sampling with time series, and particularly the `neuralforecast` implementation with the mask system is not obvious to understand. 
And if you also look at the interactions with other options such as `start_padding_enable`, you can quickly get lost :)

PR #1353, opened 2025-07-01
remove nbdev

deven367:
This PR is a WIP to remove `nbdev` from `neuralforecast`.

cc @goodwanghan @marcopeix


deven367:
tests for `AutoModels` have been moved with their respective models 
cc @marcopeix

tests/conftest.py
```diff
@@ -0,0 +1,43 @@
+import logging
+import os
+
+os.environ["PYTORCH_ENABLE_MPS_FALLBACK"] = "1" # set before neuralforecast import
```
JQGoh: I notice that in your ci.yaml file, you have specification of `PYTORCH_ENABLE_MPS_FALLBACK` as well depending whether the test run is on macOS platform
https://github.com/Nixtla/neuralforecast/pull/1353/files#diff-2500680f4bc6c1b75c3d4b36372bf4d64c5f603b90bfd7a5186f66a20329d16aR33-R39

How about we revise this line in `conftest.py` such that we will specify `PYTORCH_ENABLE_MPS_FALLBACK=1` whenever the environment is of macOS? Something like (as part of your `setup_test_environment`)

```
@pytest.fixture(scope="session", autouse=True)        
def setup_test_environment():
    """Configure the environment for all neural forecast tests."""
    if sys.platform.startswith('darwin'):
        os.environ['PYTORCH_ENABLE_MPS_FALLBACK'] = 1
        
    logging.getLogger("pytorch_lightning").setLevel(logging.ERROR)
    logging.getLogger("lightning_fabric").setLevel(logging.ERROR)
    yield        
```

In this manner, we can also revise the `ci.yaml` and omit the explicit environment variable setting.

tests/test_common/test_base_auto.py
```diff
@@ -0,0 +1,147 @@
+import logging
+import warnings
+
+import optuna
+import pandas as pd
+import pytest
+import pytorch_lightning as pl
+from fastcore.test import test_eq as _test_eq
```
JQGoh: Suggest to replace `_test_eq` with assertion tests, since we want to remove `fastcore.test`. I doubt that they work well with pytest ecosystem that tools like pytest-rerunfailures, pytest-cov, and pytest-xdist assume standard assert usage for proper behavior.

tests/test_common/test_model_checks.py
```diff
@@ -0,0 +1,83 @@
...
+    TSMixer,
+    TSMixerx,
+    VanillaTransformer,
+    iTransformer,
+)
+
+
+@pytest.fixture
+def setup_module():
+    models = [
```
JQGoh: I suggest that we use [MODEL_FILENAME_DICT](https://github.com/Nixtla/neuralforecast/blob/0a08a235e0584dad69ccef615760cc794d36fcee/neuralforecast/core.py#L139) from core.py file. Try to reduce the chance of code duplication if possible, so we can always refer to `MODEL_FILENAME_DICT` as the exhaustive list of models

If we want to omit a certain models for test purpose, we can customize based on the list of `MODEL_FILENAME_DICT` over here.

tests/test_common/test_scalers.py
```diff
@@ -0,0 +1,47 @@
...
+    x2 = 10**1 * np.arange(36)[:, None]
+
+    np_x = np.concatenate([x1, x2], axis=1)
+    np_x = np.repeat(np_x[None, :,:], repeats=2, axis=0)
+    np_x[0,:,:] = np_x[0,:,:] + 100
+
+    np_mask = np.ones(np_x.shape)
+    np_mask[:, -12:, :] = 0
+
+    print(f'x.shape [batch, time, features]={np_x.shape}')
```
JQGoh: Omit the print statement from test? Can move the setup of variables into pytest fixture

tests/test_core.py
```diff
@@ -0,0 +1,1995 @@
...
+    )
+
+
+def test_predict_insample_step_size(setup_airplane_data):
+    AirPassengersPanel_train, _ = setup_airplane_data
+
+    h = 12
+    train_end = AirPassengersPanel_train['ds'].max()
+    sizes = AirPassengersPanel_train['unique_id'].value_counts().to_numpy()
+    for step_size, test_size in [(7, 0), (9, 0), (7, 5), (9, 5)]:
```
JQGoh:
```
    for step_size, test_size in [(7, 0), (9, 0), (7, 5), (9, 5)]:
```
Consider `step_size, test_size` to be part of pytest.mark.parametrize for test, instead of for-loop.

deven367: Done

tests/test_common/test_model_checks.py
deven367: That's a good idea, I've tried to implement something similar here, [b532cc7](https://github.com/Nixtla/neuralforecast/pull/1353/commits/b532cc7bd9aa88e1d596a09f3cd29598dd767921)

tests/conftest.py
JQGoh: On a second thought and discussed, think it is also good to explicitly declare the `PYTORCH_ENABLE_MPS_FALLBACK=1` in the ci files. As you will need to declare this before the import of `neuralforecast` library

tests/test_common/test_base_auto.py
deven367: Done!

tests/test_common/test_scalers.py
deven367: Done!

deven367:
@JQGoh I'm dropping support for py3.13 in this PR as it's getting too complicated to handle it in a single PR, we'll open a new PR once this is merged to add this functionality. 

cc @marcopeix

JQGoh:
@deven367 
I did try to run the tests/test_common/test_model_checks.py tests locally (Ubuntu container env), I could complete the test run for Python3.10, but not Python3.11. 

I believe there are subtle issues that we have not identified
1. Running tests on MacOS platform
2. Running tests for Python3.11 onward, for both Windows and Ubuntu platforms

We have not really executed the model check run during the recent PRs (see the last cell of [model_checks.ipynb ](https://github.com/Nixtla/neuralforecast/blob/main/nbs/common.model_checks.ipynb)file, we have the directive `#| eval: false`). 

I suggest that we can focus on removing the nbdev first and skip the individual model check run for the time being. This issue can be revisited later.

cc: @marcopeix

tests/test_common/test_scalers.py
```diff
@@ -0,0 +1,45 @@
+import numpy as np
```
JQGoh, line 1: Suggest we can include tests from those like `Unit test for ReVIN, and its compatibility with distribution's scale decouple`?
You can check the last cell of https://github.com/Nixtla/neuralforecast/blob/main/nbs/common.scalers.ipynb file

tests/test_tsdataset.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+@pytest.fixture
+def setup_data():
+    temporal_df = generate_series(
+        n_series=1000, n_temporal_features=0, equal_ends=False
+    )
+    sorted_temporal_df = temporal_df.sort_values(["unique_id", "ds"])
+    unsorted_temporal_df = sorted_temporal_df.sample(frac=1.0)
+    dataset, indices, dates, ds = TimeSeriesDataset.from_df(df=unsorted_temporal_df)
```
JQGoh: Strictly speaking, we should place `TimeSeriesDataset.from_df` inside the test case `test_sort_df`. The test case aims to test the functionality (an action), we shall avoid including the functionality that is to be tested inside the fixture specifications.

tests/test_tsdataset.py
```diff
@@ -0,0 +1,207 @@
...
+    np.testing.assert_almost_equal(dataset_full.max_size, dataset_1.max_size)
+    np.testing.assert_almost_equal(dataset_full.indptr, dataset_1.indptr)
+
+
+# Testing trim_dataset functionality
+def test_trim_dataset():
+    n_static_features = 0
+    n_temporal_features = 2
+
+    # temporal_df = generate_series(
```
JQGoh: Nit: remove commented codes

tests/test_tsdataset.py
```diff
@@ -0,0 +1,207 @@
...
+        dataset_trimmed.temporal[
+            dataset_trimmed.indptr[50] : dataset_trimmed.indptr[51]
+        ].numpy(),
+    )
+
+
+def test_polars_integration():
+    n_static_features = 0
+    n_temporal_features = 2
+    temporal_df = generate_series(n_series=100,
```
JQGoh: This shares the same info as in line 137, can combine them into a fixture

tests/test_models/test_timellm.py
```diff
@@ -0,0 +1,4 @@
+import logging
+import warnings
+
+from neuralforecast.common._model_checks import check_model
```
JQGoh, line 4: Missing test for timellm? 

If we don't aim to test this, leave a remark explaining this.

tests/test_models/test_vanillatransformer.py
```diff
@@ -0,0 +1,34 @@
+from neuralforecast.auto import AutoVanillaTransformer, VanillaTransformer
+from neuralforecast.common._base_auto import MockTrial
+from neuralforecast.common._model_checks import check_model
+
+from .test_helpers import check_args
+
+
+def test_vanilla_transformer(suppress_warnings):
+    check_model(VanillaTransformer, ["airpassengers"])
```
JQGoh, line 9: I suggest that `check_model` with Air passengers data together with testing Auto*model using `check_args` and etc. to be moved to `tests/test_common/test_modelchecks.py` file. We should revise the tests to be similar to the style of `test_model_checks` test case

A few reasons why doing this:
 * Avoid rewriting similar/repeated codes
 * Using pytest.mark.parametrize to better understand the difference in test input/output
 * We can still keep specific tests into the individual files (like test_mlp.py file that checks `no leakage with test_size`. With your revision, we don't need to specify similar tests across different files

After your revision, I am keen to know whether we still face memory issue like the running model check with AirPassengers can still fail for Python3.11 on Ubuntu platform.
cc: @goodwanghan

tests/test_losses/test_pytorch.py
```diff
@@ -0,0 +1,148 @@
...
+    assert len(check.quantiles), 4
+
+
+# Unit tests
+# Check that default quantile is set to 0.5 at initialization
+def test_IQLoss_default_quantile():
+    check = IQLoss()
+    assert check.q, 0.5
+
+# Check that quantiles are correctly updated - prediction
```
JQGoh: Nit: combine this test to the test above to demonstrate its updated value differing from the default value. Same for the other similar test cases

tests/test_losses/test_pytorch.py
```diff
@@ -0,0 +1,148 @@
...
+# attribute is correctly instantiated
+def test_DistributionLoss_level():
+    check = DistributionLoss(distribution="Normal", level=[80, 90])
+    assert len(check.quantiles), 5
+
+def test_DistributionLoss_quantiles():
+    check = DistributionLoss(
+        distribution="Normal", quantiles=[0.0100, 0.1000, 0.5, 0.9000, 0.9900]
+    )
+    print(check.output_names)
```
JQGoh: Omit print, same for the other places that have this.

tests/test_losses/test_pytorch.py
```diff
@@ -0,0 +1,148 @@
...
+# Check that default quantile is set to 0.5 at initialization
+def test_IQLoss_default_quantile():
+    check = IQLoss()
+    assert check.q, 0.5
+
+# Check that quantiles are correctly updated - prediction
+def test_IQLoss_update_quantile():
+    check = IQLoss()
+    check.update_quantile([0.7])
+    assert check.q, 0.7
```
JQGoh: Must revise this to `assert check.q == 0.7`, same for the other assertions

tests/test_losses/test_numpy.py
```diff
@@ -0,0 +1,123 @@
...
+        mape_pytorch = mape_pytorch(y=self.y, y_hat=self.y_hat).numpy()
+        np.testing.assert_array_almost_equal(mape_numpy, mape_pytorch, decimal=6)
+
+    def test_smape(self):
+        smape_numpy   = smape(self.y, self.y_hat)
+        smape_pytorch = SMAPE()
+        smape_pytorch = smape_pytorch(self.y, self.y_hat).numpy()
+        np.testing.assert_array_almost_equal(smape_numpy, smape_pytorch, decimal=6)
+
+    # def test_mase(self):
```
JQGoh, line 73: Consider restoring these commented codes to be part of test cases. Not sure why they were commented out, can invite @marcopeix to share. Otherwise, we remove them

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+def setup():
+    uids = pd.Series(["id_0", "id_1"])
+    indptr = np.array([0, 4, 10], dtype=np.int32)
+    return uids, indptr
+
+
+def test_cutoff_deltas(setup):
+    uids, indptr = setup
+    h = 2
+    for step_size, freq, days in zip([1, 2], ["D", "W-THU"], [1, 14]):
```
JQGoh: Similar to the previous comment, avoid the for loop and we better use pytest.mark.parametrize for testing. Can revise similar test patterns for the other test cases that you found.

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+    AirPassengersPanel_train, _ = setup_airplane_data
+    models = [NHITS(h=12, input_size=24, max_steps=10)]
+    nf = NeuralForecast(models=models, freq="M")
+    nf.fit(AirPassengersPanel_train)
+    init_fcst = nf.predict()
+    init_cv = nf.cross_validation(AirPassengersPanel_train, use_init_models=True)
+    after_cv = nf.cross_validation(AirPassengersPanel_train, use_init_models=True)
+    nf.fit(AirPassengersPanel_train, use_init_models=True)
+    after_fcst = nf.predict()
+    assert (init_cv == after_cv).all().all()
```
JQGoh: cross_validation returns data frame, can consider using `pd.testing.assert_frame_equal` with `check_like=True` to check values.

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+    fcst.fit(
+        df=AirPassengersPanel_train, static_df=AirPassengersStatic, use_init_models=True
+    )
+    after_weights = fcst.models_init[0].mlp[0].weight.detach().clone()
+    assert np.allclose(initial_weights, after_weights), "init models should not be modified"
+    assert len(fcst.models[0].train_trajectories) > 0, (
+        "models stored trajectories should not be empty"
+    )
+
+def get_expected_size(df, h, test_size, step_size):
```
JQGoh: Suggest that we can group helper function like this into a standardized helper function file. Can rename the `tests/test_models/test_helpers.py` file to `tests/test_helpets.py`, and we move this to there for future developers' reference.

Same for the other similar helper functions such as `get_expected_cols`

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+            "AutoDIL",
+            "NHITSMQ-median",
+            "NHITSMQ-lo-80",
+            "NHITSMQ-hi-80",
+            "MyRNN",
+            "StemMulti",
+            "NHITS",
+        ]
+
+# # Unit test for core/model interactions
```
JQGoh: Can remove this. The airPassengers test can cover this.

Similarly， remove other non-used commented codes such as line 823

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+def test_cross_validation(setup_test_environment, h=12, test_size=12):
+    df = AirPassengersPanel
+    static_df = AirPassengersStatic
+    if (test_size - h) % 1:
+        raise Exception("`test_size - h` should be module `step_size`")
+
+    n_windows = int((test_size - h) / 1) + 1
+    Y_test_df = df.groupby("unique_id").tail(test_size)
+    Y_train_df = df.drop(Y_test_df.index)
+    config = {
```
JQGoh, line 701: Moved these commonly used fixtures to the contest file or to the top?

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+        fcst2 = NeuralForecast.load(path=path)
+        forecasts2 = fcst2.predict(futr_df=AirPassengersPanel_test, level=[50])
+        pd.testing.assert_frame_equal(forecasts1, forecasts2[forecasts1.columns])
+
+
+# test save and load without dataset
+def test_save_load_no_dataset(setup_airplane_data):
+    AirPassengersPanel_train, AirPassengersPanel_test = setup_airplane_data
+
+    # shutil.rmtree("examples/debug_run")
```
JQGoh: I think this is important as part of the teardown process for the test case `test_save_load`. Revise this to help with releasing resources

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+    valid_losses = nf.models[0].valid_trajectories
+    assert valid_losses[-1][1] < 40, "Validation loss is too high"
+    assert valid_losses[-1][1] > 10, "Validation loss is too low"
+
+
+def test_neural_forecast_missing_variables(setup_airplane_data):
+    """Test that fit fails appropriately when variables are not in dataframe."""
+    AirPassengersPanel_train, _ = setup_airplane_data
+
+    # Test Base Windows models with missing historical exogenous variables
```
JQGoh: 1. I first introduced the test cases for different Base model classes such as Recurrent and etc., because the model classes were not standardized. But since this https://github.com/Nixtla/neuralforecast/pull/1023, they are now sharing the same Base class. Can test NHITS model will do without the need of testing RNN, StemGNN

2. Consider pytest.mark.parametrize to group different test cases, because essentially they test different kinds of assertion errors due to fit()

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+            "airline2": [1, 0],
+        }
+    )
+
+    # Set up column renaming for Polars
+    renamer = {"unique_id": "uid", "ds": "time", "y": "target"}
+    inverse_renamer = {v: k for k, v in renamer.items()}
+
+    # Create Polars dataframes
+    AirPassengers_pl = polars.from_pandas(AirPassengersPanel_train)
```
JQGoh: I thought you already have `setup_airplane_data_polars` fixture? Reuse already defined fixture whenever we can

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+    preds_pl = nf_polars.predict()
+    cv_res_pl = nf_polars.cross_validation(
+        df=AirPassengers_pl,
+        static_df=AirPassengersStatic_pl,
+        id_col="uid",
+        time_col="time",
+        target_col="target",
+    )
+
+    def assert_equal_dfs(pandas_df, polars_df):
```
JQGoh: Moved to standardized helper function file

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+                ["cutoff", "len"]
+            ),
+            check_row_order=False,
+        )
+
+
+# Test if any of the inputs contains NaNs with available_mask = 1, fit shall raise error
+# input type is pandas.DataFrame
+# available_mask is explicitly given
+def test_masks_polars():
```
JQGoh: This test name implies it is polars related test. But there is no polars involved right?

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+
+def test_neural_forecast_invalid_optimizer():
+    """Test that invalid optimizers raise appropriate exceptions for different model types."""
+
+    # Test that if the user-defined optimizer is not a subclass of torch.optim.Optimizer,
+    # it fails with exception. Tests cover different types of base classes such as
+    # BaseWindows, BaseRecurrent, BaseMultivariate
+
+    # Test BaseWindows model (NHITS)
+    with pytest.raises(Exception) as exc_info:
```
JQGoh, line 1382: Consider pytest.mark.parametrize for these cases. Same for `test_customized_behavior` and other similar test cases

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+            nf.fit(AirPassengersPanel_train)
+            assert any(
+                "ignoring lr_scheduler_kwargs as the lr_scheduler is not specified"
+                in str(w.message)
+                for w in issued_warnings
+            ), f"Expected lr_scheduler_kwargs warning not found for {nf_model.__name__}"
+
+# # test conformal prediction, method=conformal_distribution
+
+# prediction_intervals = PredictionIntervals()
```
JQGoh: Why conformal prediction tests are omitted?

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+
+        params = {"h": 12, "input_size": 24, "max_steps": 1, "loss": IQLoss()}
+        if nf_model.__name__ == "TSMixer":
+            params.update({"n_series": 2})
+        models.append(nf_model(**params))
+
+    nf = NeuralForecast(models=models, freq="M")
+    nf.fit(AirPassengersPanel_train, prediction_intervals=prediction_intervals)
+
+    # Test default prediction
```
JQGoh, line 1617: Pytest.mark.parametrize these few tests?

tests/test_losses/test_pytorch.py
deven367: Done!

tests/test_losses/test_pytorch.py
deven367: Done!

tests/test_core.py
deven367: Done!

tests/test_core.py
deven367: Done!

tests/test_core.py
deven367: Done!

tests/test_core.py
deven367: Done!

tests/test_core.py
deven367: Done!

tests/test_models/test_timellm.py
marcopeix: I think it's okay not to test TimeLLM for now. It needs to load an LLM, it's hard to fit in memory, it's pretty slow, so let's skip it for now.

tests/test_core.py
```diff
@@ -0,0 +1,1994 @@
...
+    fcst.fit(AirPassengersPanel_train)
+    last_log = f"lightning_logs/{os.listdir('lightning_logs')[-1]}"
+    no_chkpt_found = ~np.any(
+        [file.endswith("checkpoints") for file in os.listdir(last_log)]
+    )
+    assert no_chkpt_found ==  False
+
+# test `enable_checkpointing=False` should not generate chkpt
+def test_no_checkpointing(setup_airplane_data):
+    AirPassengersPanel_train, _ = setup_airplane_data
```
marcopeix, line 984: This line is repeated many times. Maybe we can run it once at the top and just pass train and test as a parameter to the functions?

tests/test_losses/test_numpy.py
marcopeix: We discussed some time ago to remove numpy losses from neuralforecast. Let me confirm that.

marcopeix:
For the memory issues on MacOS, despite setting the MPS fallback variable, maybe we can try running the tests on a single thread? Don't know if they are running in parallel or not, but we could try setting `n_workers=1` see if it solves the issue.

tests/test_models/test_timellm.py
deven367: Done!

tests/test_core.py
deven367: Done!

tests/test_core.py
deven367: I've included those tests now!

tests/test_tsdataset.py
deven367: Done!

tests/test_tsdataset.py
deven367: Done!

tests/test_losses/test_pytorch.py
deven367: Done

tests/test_core.py
deven367: done!

deven367:
@marcopeix I just checked, `pytest` is single threaded by default

tests/test_common/test_scalers.py
deven367: they've now been added! [2191e09](https://github.com/Nixtla/neuralforecast/pull/1353/commits/2191e09d5fc429b0dd95d681185252f2ef301c70)

tests/test_models/test_vanillatransformer.py
deven367: I believe that the original plan was to have `check_model(model, ["airpassengers", "losses"])` together, I can't remember the reason why we kept them separate, @marcopeix can you recall this discussion?

tests/test_core.py
deven367: This is resolved!

tests/test_core.py
deven367: Yes, it is now resolved!

tests/test_core.py
deven367: It's a pytest fixture, maybe we can make it better in the next PR

tests/test_core.py
deven367: The test was misnamed, it is now fixed!

tests/test_core.py
deven367: 1. Maybe we can then address in another PR, I didn't realize that they had the same base class
2. The tests are now parameterized!

tests/test_tsdataset.py
deven367: Done in [21acb02](https://github.com/Nixtla/neuralforecast/pull/1353/commits/21acb02b4b0fd791341ac64f00db028c0e6a05ea)

tests/test_losses/test_numpy.py
deven367: @JQGoh We're planning to handle this in a separate PR, it needs a discussion with @elephaint and @cchallu 
cc @marcopeix

tests/test_models/test_vanillatransformer.py
deven367: The main reason to keep them separate was to have smaller tests, more contained, easier to maintain! however we can circle back to this in the next PR

JQGoh: > The main reason to keep them separate was to have smaller tests, more contained, easier to maintain! however we can circle back to this in the next PR

I think it makes sense to separate the test on 'losses' vs 'air_passengers'. However, I have a different opinion on wiring similar test purposes for different models, with only change to the parameters like model choice or config. It is best we consolidate them into a single file and parametrize, and avoid introducing similar/equivalent code logics to different files. This is harder to maintain for the future developers as they need to read different files and come to the conclusion that we are testing the same purpose. 

I won't suggest to include these new changes in this PR, best to consolidate them now than later. 

I can help to demonstrate/show more on the consolidation ideas next week, if you need more inputs.

tests/test_core.py
JQGoh: > 1. Maybe we can then address in another PR, I didn't realize that they had the same base class
> 2. The tests are now parameterized!

Thanks for parametrizing. However, this is not needed for test purpose as one model used for test like NHITS is good enough. We can remove the other models for test in the future PR

tests/test_models/test_vanillatransformer.py
deven367: How about something like this?
```py
@pytest.mark.parametrize("model", [v for k,v in MODEL_FILENAME_DICT.items() if 'auto' not in k])
@pytest.mark.parametrize("check", ["losses", "airpassengers"])
def test_model_checks(model, check):
    check_model(model, checks=[check])
    check_args(model, exclude_args=['cls_model'])
```

goodwanghan: I don't agree with this:

> It is best we consolidate them into a single file and parametrize

But I agree with this

> avoid introducing similar/equivalent code logics to different files.

The reason not to put all models into a single file with parameterization:

1. It will be hard to handle exceptions, what if a model only pass on "losses" but it is an expected behavior? We will have a lot of if else inside the general test function.
2. All models must be imported in order to start running the test, with one failure, it will completely fail
3. Each model still has its own special settings and tests, so a separate test file is often inevitable, taking AutoBiTCN, BiTCN as an example, if we just created these two models, do we want to put the relevant tests in one file, or do we want to add part of the tests to a centralized place and add the rest in a new test file?
4. The parameterized test and its file will often cause merge conflicts

The ideal way should be creating a model test suite, extracting common logic as much as possible into the base test class, and this allows overriding too. And we should still keep each model test in separate files. Each file may have just a few lines of code, but the separation is very important, a little bit redundancy is fine.

tests/test_core.py
deven367: Done!

tests/test_models/test_vanillatransformer.py
goodwanghan: Notice this PR doesn't have coverage number. I suggest we keep it as is for now. And refactor it in the next PR (when coverage number is added)

marcopeix (approved):

tests/test_models/test_vanillatransformer.py
JQGoh: > How about something like this?
> 
> ```python
> @pytest.mark.parametrize("model", [v for k,v in MODEL_FILENAME_DICT.items() if 'auto' not in k])
> @pytest.mark.parametrize("check", ["losses", "airpassengers"])
> def test_model_checks(model, check):
>     check_model(model, checks=[check])
>     check_args(model, exclude_args=['cls_model'])
> ```

Yes, that was my thought but depending on test behavior, we don't necessarily group the `losses` and `airpassengers` as parametrize parametrrs

JQGoh: > I don't agree with this:
> 
> > It is best we consolidate them into a single file and parametrize
> 
> But I agree with this
> 
> > avoid introducing similar/equivalent code logics to different files.
> 
> The reason not to put all models into a single file with parameterization:
> 
> 1. It will be hard to handle exceptions, what if a model only pass on "losses" but it is an expected behavior? We will have a lot of if else inside the general test function.
> 2. All models must be imported in order to start running the test, with one failure, it will completely fail
> 3. Each model still has its own special settings and tests, so a separate test file is often inevitable, taking AutoBiTCN, BiTCN as an example, if we just created these two models, do we want to put the relevant tests in one file, or do we want to add part of the tests to a centralized place and add the rest in a new test file?
> 4. The parameterized test and its file will often cause merge conflicts

Understand your points on this 👍 At first I suggested the parametrization is based on the current test requirement/need. Especially when we have many model specific test logic, we separate them. 

> The ideal way should be creating a model test suite, extracting common logic as much as possible into the base test class, and this allows overriding too. And we should still keep each model test in separate files. Each file may have just a few lines of code, but the separation is very important, a little bit redundancy is fine.

Yes, we can work on this in the future

PR #1360, opened 2025-07-10
rm unused `y_insample` arg from pytorch losses

deven367:
This PR removes the `y_insample` arg from pytorch losses of 
1. MAE
2. MSE
3. RMSE
4. MAPE
5. SMAPE
6. MQLoss
7. TukeyLoss
8. HuberQLoss
9. HuberIQLoss

cc @marcopeix


elephaint:
You can't remove that argument, it's intentionally there to keep api consistent for losses which is necessary in base_model

deven367:
Ah, I see, it was leading to failing tests with `pytest`, I'll set it to `None` then, how's that?

PR #1362, opened 2025-07-14
Dev environment installation using uv

JQGoh:
Suggest the installation of dev environment using [astral's uv](https://github.com/astral-sh/uv)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,13 +36,12 @@ Bug fixes and features are added through pull requests (PRs).
 Create a virtual environment to install the library's dependencies. We recommend [astral's uv](https://github.com/astral-sh/uv).
 Once you've created the virtual environment you should activate it and then install the library in editable mode along with its development dependencies.
 
-Install cpu-only version on Linux OS:
- * The added `--extra-index-url` follows the instructions provided by https://pytorch.org/get-started/locally/ for individual OS platform.
+Install cpu-only version:
 ```bash
 pip install uv
 uv venv --python 3.10
 source .venv/bin/activate
-uv pip install -Ue .[dev] --extra-index-url https://download.pytorch.org/whl/cpu
+uv pip install -Ue .[dev]
 ```
 
 Installation with cuda support


CONTRIBUTING.md
```diff
@@ -32,22 +32,27 @@ Bug fixes and features are added through pull requests (PRs).
...
-```
-conda env update -f environment-cpu.yml  # choose this if you want to install the CPU-only version of neuralforecast
-conda env update -f environment-cuda.yml # choose this if you want to install the CUDA-enabled version of neuralforecast
+
+Installation with cuda support
+```bash
+pip install uv
+uv venv --python 3.10
+source .venv/bin/activate
+uv pip install -Ue .[dev,cuda]
```
JQGoh, line 52: @marcopeix 
1. Perhaps there are good reasons that the tests in [.circleci/config.yml](https://github.com/Nixtla/neuralforecast/blob/main/.circleci/config.yml) are still using `micromamba`. If this work make sense that we can migrate to using `uv` package manager, I could help to update the .circleci/config.yml file.
2. I don't have access to CUDA resource. But based on my understanding, this requires `pytorch-cuda>=11.8` to setup. If you could help to test this, appreciate your help with this.
3. Let me know what do you think about this PR on migrating to `uv` package manager. Most recently I gave this a try and I believe it could be useful for the future developers.

marcopeix: 1. I think we can move to uv in CircleCI. See [here](https://github.com/Nixtla/hierarchicalforecast/blob/main/.circleci/config.yml) for an example.
2. Are you talking about the tests in CircleCI? We don't use GPUs in the test, we use the cpu dependencies.
3. Sure, migrating to uv makes sense, we're doing for other packages as well!

Thanks a lot!

JQGoh: @marcopeix 
Regarding 2: nope, I am not talking about tests in CirclrCI. I am referring to the changes in this PR whether are the revisions valid for users who want to install with CUDA dependency. I am asking for help in reviewing PR and best to validate the proposal.

JQGoh: @marcopeix This is ready for review

nbs/docs/getting-started/04_installation.ipynb
```diff
@@ -57,20 +57,22 @@
     "$ git clone https://github.com/Nixtla/neuralforecast.git && cd neuralforecast\n",
```
marcopeix, line 57: <p>To stay consistent with the previous instructions, could we keep the name of the env as neuralforecast?</p><p> </p><p>So, it would be:</p><p> </p><p>uv venv neuralforecast --python 3.10</p><p>source neuralforecast/bin/activate</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/neuralforecast/pull/1362/discussion/#comment-2214177011'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='1'/>

CONTRIBUTING.md
```diff
@@ -32,22 +32,26 @@ Bug fixes and features are added through pull requests (PRs).
...
+#### Set up an environment
+Create a virtual environment to install the library's dependencies. We recommend [astral's uv](https://github.com/astral-sh/uv).
+Once you've created the virtual environment you should activate it and then install the library in editable mode along with its development dependencies.
+
+Install cpu-only version:
+```bash
+pip install uv
+uv venv --python 3.10
+source .venv/bin/activate
+uv pip install -Ue .[dev]
```
marcopeix, line 44: Same here, can we keep the name of the env as neuralforecast?

CONTRIBUTING.md
```diff
@@ -32,22 +32,26 @@ Bug fixes and features are added through pull requests (PRs).
...
-```
-conda env update -f environment-cpu.yml  # choose this if you want to install the CPU-only version of neuralforecast
-conda env update -f environment-cuda.yml # choose this if you want to install the CUDA-enabled version of neuralforecast
+
+Installation with cuda support
+```bash
+pip install uv
+uv venv --python 3.10
+source .venv/bin/activate
+uv pip install -Ue .[dev,cuda]
```
marcopeix, line 52: And here too, keep neuralforecast as the name of the env

marcopeix (changes requested):

CONTRIBUTING.md
JQGoh: @marcopeix 
I suggest that we keep the command `uv venv --python 3.10` as it is. 

When we did `uv venv --python 3.10`, this will create a hidden folder `.venv` which does not conflict with a created directory name.

However, `uv venv neuralforecast --python 3.10`, this results in
```
root@5791ba15c061:~/retest# uv venv neuralforecast --python 3.10
Using CPython 3.10.12 interpreter at: /usr/bin/python3.10
Creating virtual environment at: neuralforecast
uv::venv::creation

  × Failed to create virtualenv
  ╰─▶ The directory `neuralforecast` exists, but it's not a virtual environment
```
which has a conflict with the existing folder name = `neuralforecast`. `uv` expects to **create a folder instead creating a hidden folder**.


The original instruction also aligns with the contributing guide as detailed in
https://github.com/Nixtla/nixtla/blob/main/CONTRIBUTING.md

PS: We can also give another name "neural-forecast" but I found similar folder names created also cause unnecessary confusion.

marcopeix: Sounds good, I'll dismiss this then!

marcopeix (approved):
Looks good! Tested on my end and seems to work well.

elephaint:
@JQGoh Did you check that uv also installs CUDA when on Windows? Because that's what would prevent using uv with NF. That's why we also need the separate cpu and cuda yamls.

Also the changes to the guides completely ignore Windows users (the suggested code isn't applicable for them).

PR #1364, opened 2025-07-21
[FIX] Fixes for using uv

elephaint:
Fixes some bugs of #1362:
- Correct syntax for Windows
- Removes unnecessary added dependencies
- Restores numpy version requirement


settings.ini
```diff
@@ -15,10 +15,10 @@ language = English
 custom_sidebar = True
 license = apache2
 status = 2
-requirements = coreforecast>=0.0.6 fsspec numpy<2 pandas>=1.3.5 torch>=2.0.0,<=2.6.0 pytorch-lightning>=2.0.0 ray[tune]>=2.2.0 optuna utilsforecast>=0.2.3 scipy
+requirements = coreforecast>=0.0.6 fsspec numpy>=1.21.6 pandas>=1.3.5 torch>=2.0.0,<=2.6.0 pytorch-lightning>=2.0.0 ray[tune]>=2.2.0 optuna utilsforecast>=0.2.3
```
elephaint: @JQGoh @marcopeix I didn't see any reason why these requirements were changed in #1362, so I'm reverting them back. Please let me know if I'm wrong here.

settings.ini
```diff
@@ -15,10 +15,10 @@ language = English
...
 spark_requirements = fugue pyspark>=3.5
 aws_requirements = fsspec[s3]
-dev_requirements = black fastcore<=1.7.29 gitpython hyperopt ipython<=8.32.0 matplotlib mypy nbdev==2.3.25 polars pre-commit pyarrow ruff s3fs transformers jupyterlab numba ipykernel pip
+dev_requirements = black fastcore<=1.7.29 gitpython hyperopt ipython<=8.32.0 matplotlib mypy nbdev==2.3.25 polars pre-commit pyarrow ruff s3fs transformers
```
elephaint, line 21: @JQGoh @marcopeix I didn't see any reason why these requirements were changed in #1362, so I'm reverting them back. Please let me know if I'm wrong here.

CONTRIBUTING.md
```diff
@@ -36,20 +36,26 @@ Bug fixes and features are added through pull requests (PRs).
...
-uv pip install -Ue .[dev,cuda]
+uv pip install -Ue .[dev] # cpu-only version
+uv pip install -Ue .[dev, cuda] # cuda version
+```
+
+- On Windows:
+```bash
+.\.venv\Scripts\activate
+uv pip install -Ue .[dev] # cpu-only version
+uv pip install -Ue .[dev, cuda] # cuda version
```
elephaint: @JQGoh Can you verify that this works under Windows?

settings.ini
JQGoh: @elephaint 
I chose `numpy<2` because I noticed that the [ci pipeline](https://github.com/Nixtla/neuralforecast/blob/30e548fa53ad7e0fa3dd11bb4b1b8eea91f18dba/.github/workflows/ci.yaml#L40) is pinned to this. We can revert it back to numpy>=1.21.6

Regarding the scipy library, I noticed that nbeats model [does import BSpline from scipy](https://github.com/Nixtla/neuralforecast/blob/30e548fa53ad7e0fa3dd11bb4b1b8eea91f18dba/neuralforecast/models/nbeats.py#L15).
Think when I first tried this installation, scipy was not installed and it failed.

settings.ini
JQGoh: @elephaint The reason I chose `jupyterlab numba ipykernel pip` as part of considerations was align with dependencies mentioned in the [old environment-cpu.yml](https://github.com/Nixtla/neuralforecast/pull/1362/files#diff-6d807954b0c96c01bea7435833a15afa6c011eb97cbe404ebfdc5b81af4e3a50) file, for instance the mention of [jupyterlab](https://github.com/Nixtla/neuralforecast/pull/1362/files#diff-6d807954b0c96c01bea7435833a15afa6c011eb97cbe404ebfdc5b81af4e3a50L11)

CONTRIBUTING.md
JQGoh: Thanks for your reminder. Indeed I overlooked the Windows platform requirement.

settings.ini
elephaint: Thanks!

Scipy is only necessary for non-default arguments in NBEATS, so I'm fine leaving it out to keep the lib smaller.

Good point about the CI, I'll see if it can be increased there to be consistent. Small note: when pinning something with a minimum version, make sure to keep the minimum version, i.e. `numpy>=1.21.6,<2` (covering the edge case where someone has `numpy<1.21.6`)

CONTRIBUTING.md
elephaint: It doesn't work, I tried

settings.ini
```diff
@@ -15,10 +15,10 @@ language = English
 custom_sidebar = True
 license = apache2
 status = 2
-requirements = coreforecast>=0.0.6 fsspec numpy<2 pandas>=1.3.5 torch>=2.0.0,<=2.6.0 pytorch-lightning>=2.0.0 ray[tune]>=2.2.0 optuna utilsforecast>=0.2.3 scipy
+requirements = coreforecast>=0.0.6 fsspec numpy>=1.21.6 pandas>=1.3.5 torch>=2.4.0 pytorch-lightning>=2.0.0 ray[tune]>=2.2.0 optuna utilsforecast>=0.2.3
 spark_requirements = fugue pyspark>=3.5
 aws_requirements = fsspec[s3]
-dev_requirements = black fastcore<=1.7.29 gitpython hyperopt ipython<=8.32.0 matplotlib mypy nbdev==2.3.25 polars pre-commit pyarrow ruff s3fs transformers jupyterlab numba ipykernel pip
+dev_requirements = black fastcore<=1.7.29 gitpython hyperopt ipython<=8.32.0 matplotlib mypy nbdev==2.3.25 polars pre-commit pyarrow ruff s3fs transformers
 cuda_requirements = pytorch-cuda>=11.8
```
JQGoh: Thanks for the fix. Should we still keep `cuda_requirements = pytorch-cuda>=11.8`, since this is removed in setup.py file?

settings.ini
elephaint: Ah, makes sense! I think that was a bit of a relic from the past. The issue with adding these deps to dev in settings.ini is that we incur the dependency penalty from them across all CI tests then (in the past we didn't get that, because the yaml files weren't actually being used in testing). For that reason, I think it's better to leave it out out settings.ini.

settings.ini
elephaint: Good point! I'll remove it

marcopeix (approved):

PR #1368, opened 2025-08-14
[FEAT] Any horizon prediction

elephaint:
Adds possibility to make predictions longer than the trained horizon to `.predict` and `.cross_validation`

Todo:
- [x] Any horizon prediction
- [x] Any horizon cross-validation (only with `refit=True`)
- [ ] Test for various model types
- [x] Test cross-validation
- [ ] Add unit tests
- [x] Add futr_exog
- [x] ~~Add hist_exog.~~ Skipping this for now, this will require significant API changes (need to add a "futr_hist_df" argument to predict, for example, which I think is very confusing.)

Checks:
- [x] ~~Check requirement for futr_df in predict~~
- [ ] Check multivariate
- [ ] Check distributional losses
- [ ] Check MQlosses
- [ ] Check IQLoss
- [ ] Check all predict arguments
- [ ] Check all cross-validation arguments
- [ ] Check recurrent models

The following code works:

```
import pandas as pd

from neuralforecast.models import NLinear
from neuralforecast.losses.pytorch import DistributionLoss, MAE
from neuralforecast.utils import AirPassengersPanel, AirPassengersStatic, augment_calendar_df

AirPassengersPanel, calendar_cols = augment_calendar_df(df=AirPassengersPanel, freq='M')
train_h = 12
h = 23

Y_train_df = AirPassengersPanel[AirPassengersPanel.ds<AirPassengersPanel['ds'].values[-h]] # 132 train
Y_test_df = AirPassengersPanel[AirPassengersPanel.ds>=AirPassengersPanel['ds'].values[-h]].reset_index(drop=True) # 12 test

model = NLinear(h=train_h,
                 input_size=24,
                 loss=DistributionLoss(distribution='StudentT', level=[80, 90], return_params=True),
                # loss=MAE(),
                 scaler_type='robust',
                 learning_rate=1e-3,
                 max_steps=200,
                 val_check_steps=50,
                 early_stop_patience_steps=2)

nf = NeuralForecast(
    models=[model],
    freq='ME'
)
nf.fit(df=Y_train_df, static_df=AirPassengersStatic, val_size=train_h)
forecasts = nf.predict(h=h)

forecasts = nf.cross_validation(df=Y_train_df, n_windows=2, refit=True)
```

### DIFF
--- a/tests/test_losses/test_iqloss.py
+++ b/tests/test_losses/test_iqloss.py
@@ -1,13 +1,22 @@
 # %% Test IQLoss for all types of architectures
 import matplotlib.pyplot as plt
 import pandas as pd
+import pytest
+import sys
+
 from neuralforecast import NeuralForecast
 from neuralforecast.common.enums import TimeSeriesDatasetEnum
 from neuralforecast.models import NBEATSx, NHITS, TSMixerx, LSTM, BiTCN
 from neuralforecast.losses.pytorch import IQLoss
 from tests.helpers.data import air_passengers
 
-
+@pytest.mark.skipif(
+    sys.platform == "darwin",
+    reason=(
+        "RuntimeError: MPS backend out of memory (MPS allocated: 8.00 MB, other allocations: 16.00 KB). Tried to allocate 256 bytes on shared pool."
+        "Test failed in MacOS as shown by https://github.com/Nixtla/neuralforecast/actions/runs/17818711510/job/50656840114?pr=1368"
+    )
+)
 def test_iqloss(show_plot=False):
     Y_train_df, Y_test_df, _, AirPassengersStatic = air_passengers(h=12)
 
--- a/tests/test_losses/test_isqfdistribution.py
+++ b/tests/test_losses/test_isqfdistribution.py
@@ -1,6 +1,8 @@
 # %% Test ISQF DistributionLoss for all types of architectures
 import matplotlib.pyplot as plt
 import pandas as pd
+import pytest
+import sys
 
 from neuralforecast import NeuralForecast
 from neuralforecast.common.enums import TimeSeriesDatasetEnum
@@ -8,7 +10,13 @@ from neuralforecast.models import NHITS, BiTCN
 from neuralforecast.losses.pytorch import DistributionLoss
 from tests.helpers.data import air_passengers
 
-
+@pytest.mark.skipif(
+    sys.platform == "darwin",
+    reason=(
+        "RuntimeError: MPS backend out of memory (MPS allocated: 8.00 MB, other allocations: 16.00 KB). Tried to allocate 256 bytes on shared pool."
+        "Test failed in MacOS as shown by https://github.com/Nixtla/neuralforecast/actions/runs/17818711510/job/50656840114?pr=1368"
+    )
+)
 def test_isqf_distribution_loss(show_plot=False):
     Y_train_df, Y_test_df, _, AirPassengersStatic = air_passengers(h=12)
     max_steps = 10


JQGoh:
* We may need to handle the case of forecast with conformal prediction intervals whereby predict_h > train_h.  [neuralforecast/core.py](https://github.com/Nixtla/neuralforecast/blob/b427ef1306c03731c22bac3d4e2272adb58d66d3/neuralforecast/core.py#L1825) specifies self.h meanwhile model.predict specifed with [given h](https://github.com/Nixtla/neuralforecast/blob/b427ef1306c03731c22bac3d4e2272adb58d66d3/neuralforecast/core.py#L1825). Note also the conformity scores are only calculated during fit, perhaps we can enforce this is available only upon refit, whereby for predict_h = train_h  (validation preventing predict_h > train_h when conformal interval is expected)?
  * Shall we provide h parameter for BaseAuto class such that the auto model can perform forecast with larger horizon too? https://github.com/Nixtla/neuralforecast/blob/b427ef1306c03731c22bac3d4e2272adb58d66d3/neuralforecast/common/_base_auto.py#L448

elephaint:
> * We may need to handle the case of forecast with conformal prediction intervals whereby predict_h > train_h.  [neuralforecast/core.py](https://github.com/Nixtla/neuralforecast/blob/b427ef1306c03731c22bac3d4e2272adb58d66d3/neuralforecast/core.py#L1825) specifies self.h meanwhile model.predict specifed with [given h](https://github.com/Nixtla/neuralforecast/blob/b427ef1306c03731c22bac3d4e2272adb58d66d3/neuralforecast/core.py#L1825). Note also the conformity scores are only calculated during fit, perhaps we can enforce this is available only upon refit, whereby for predict_h = train_h  (validation preventing predict_h > train_h when conformal interval is expected)?
> 
>     * Shall we provide h parameter for BaseAuto class such that the auto model can perform forecast with larger horizon too? https://github.com/Nixtla/neuralforecast/blob/b427ef1306c03731c22bac3d4e2272adb58d66d3/neuralforecast/common/_base_auto.py#L448

Good points!
- I'd indeed enforce only available upon refit.
- No opinion; not sure how much work it is to implement. Wdyt? If it's too much work I'd vote to not do it for now and leave it for future.

JQGoh:
> > * We may need to handle the case of forecast with conformal prediction intervals whereby predict_h > train_h.  [neuralforecast/core.py](https://github.com/Nixtla/neuralforecast/blob/b427ef1306c03731c22bac3d4e2272adb58d66d3/neuralforecast/core.py#L1825) specifies self.h meanwhile model.predict specifed with [given h](https://github.com/Nixtla/neuralforecast/blob/b427ef1306c03731c22bac3d4e2272adb58d66d3/neuralforecast/core.py#L1825). Note also the conformity scores are only calculated during fit, perhaps we can enforce this is available only upon refit, whereby for predict_h = train_h  (validation preventing predict_h > train_h when conformal interval is expected)?
> >   
> >   * Shall we provide h parameter for BaseAuto class such that the auto model can perform forecast with larger horizon too? https://github.com/Nixtla/neuralforecast/blob/b427ef1306c03731c22bac3d4e2272adb58d66d3/neuralforecast/common/_base_auto.py#L448
> 
> Good points!
> 
> * I'd indeed enforce only available upon refit.
> * No opinion; not sure how much work it is to implement. Wdyt? If it's too much work I'd vote to not do it for now and leave it for future.

Sounds good. I am still working on the basic setup etc. I have a hunch that the support for "AutoModel can perform forecast with larger horizon" does not need significant work. I will know more once I get the basic tests working.

tests/backward_comp/test_backward_comp.py
```diff
@@ -0,0 +1,56 @@
...
+        (
+            DeepAR,
+            {"h": 5, "input_size": 12, "max_steps": 2, "futr_exog_list": ["trend"]},
+        ),
+        (NLinear, {"h": 5, "input_size": 12, "max_steps": 2}),
+        (TSMixer, {"h": 5, "input_size": 12, "n_series": 2, "max_steps": 2}),
+    ],
+)
+def test_backward_comptability(model, kwargs, save_model=False):
+    # Fixture generated with save_model=True, on codespace env (Ubuntu)
```
JQGoh, line 33: Note that the generated fixture files in this PR were generated based on the commit https://github.com/Nixtla/neuralforecast/commit/29769c7be989764bde5be76025bbeca35c5c9665,
which does not include code logics introduced in this PR. 

This is to support the future tests that we prevent issue like https://github.com/Nixtla/neuralforecast/pull/1224 from happening again

tests/backward_comp/test_backward_comp.py
```diff
@@ -0,0 +1,56 @@
...
+    if save_model:
+        nf = NeuralForecast(
+            models=[model(**kwargs)],
+            freq="ME",
+        )
+
+        nf.fit(df=train_df)
+        nf.predict(futr_df=test_df)
+        os.makedirs(save_path, exist_ok=True)
+        nf.save(path=save_path, model_index=None, overwrite=True, save_dataset=True)
```
marcopeix: Since the models were previously fitted on `train_df`, we could save them with `save_dataset=False` and just pass `train_df` in `predict` as the input dataframe to get predictions. That way, we don't save the same dataset three times. Would that work?

.circleci/config.yml
```diff
@@ -18,7 +18,7 @@ jobs:
             source .venv/bin/activate
             uv pip install ".[dev]"            
             export GIT_PYTHON_REFRESH=quiet
-            pytest
+            pytest -k "not test_autonlinear_longer_horizon"
```
marcopeix, line 21: Question: why do we skip this test?

tests/backward_comp/test_backward_comp.py
JQGoh: Yes, that is much better. I also thought of creating a list of three models directly but I think your suggestion is better

.circleci/config.yml
JQGoh: This was the test that resulted in the stuck case that I mentioned in Slack channel. Running the tests on Github action works fine and I suggest that we can create a separate issue to track this.

neuralforecast/common/_base_model.py
```diff
@@ -310,6 +320,7 @@ def __init__(
         self.input_size = input_size
         self.windows_batch_size = windows_batch_size
         self.start_padding_enabled = start_padding_enabled
+        self.predict_horizon = None  # Used in recurrent prediction whereby predict h > h_train
```
elephaint: This should be `h` as default argument, otherwise we always need to check for if it exists.

neuralforecast/common/_base_model.py
```diff
@@ -1124,7 +1135,7 @@ def _predict_step_recurrent_batch(
         # Initialize results array
         n_outputs = len(self.loss.output_names)
         y_hat = torch.zeros(
-            (insample_y.shape[0], self.horizon_backup, self.n_series, n_outputs),
+            (insample_y.shape[0], self.predict_horizon, self.n_series, n_outputs),
```
elephaint, line 1138: How does this work if `predict_horizon` is None?

neuralforecast/common/_base_model.py
```diff
@@ -1152,7 +1163,7 @@ def _predict_step_recurrent_batch(
         )
 
         # Horizon prediction recursively
-        for tau in range(1, self.horizon_backup):
+        for tau in range(1, self.predict_horizon):
```
elephaint, line 1166: Same, `predict_horizon` can't be `None`

neuralforecast/common/_base_model.py
```diff
@@ -1271,6 +1282,169 @@ def _predict_step_direct_batch(
...
+                y_hat_median = y_hat
+                if median_idx:
+                    y_hat_median = y_hat[..., median_idx]
+                if i < self.n_predicts - 1:
+                    # Update temporal of the batch with predictions
+                    temporal = batch["temporal"]
+                    if self.MULTIVARIATE:
+                        y_hat_median = y_hat_median.swapaxes(0, 2)
+                        y_hat_median = y_hat_median.swapaxes(1, 2)
+                        y_hat_median = y_hat_median.squeeze()
```
elephaint: We can't use squeeze() because of the edge case where n_series=1 and someone uses a multivariate model. So, you have to supply the dimension to be squeezed.

neuralforecast/core.py
```diff
@@ -1093,16 +1138,39 @@ def cross_validation(
...
 
         Returns:
             fcsts_df (pandas or polars DataFrame): DataFrame with insample `models` columns for point predictions and probabilistic
                 predictions for all fitted `models`.
         """
-        h = self.h
+        if h is not None:
+            if h > self.h:
+                for model in self.models:
+                    if model.hist_exog_list:
```
elephaint, line 1159: Why are we not implementing the prediction_intervals check here?

tests/dummy/dummy_models.py
```diff
@@ -0,0 +1,319 @@
+__all__ = ["DummyUnivariate", "DummyMultivariate", "DummyRecurrent"]
```
elephaint, line 1: Nice!!

tests/helpers/__ini__.py
elephaint, line 1: name should be `__init__.py` (it missed the t)?

tests/dummy/dummy_models.py
```diff
@@ -0,0 +1,319 @@
...
+        h: int,
+        input_size: int = -1,
+        stat_exog_list: Union[List, None] = None,
+        hist_exog_list: Union[List, None] = None,
+        futr_exog_list: Union[List, None] = None,
+        inference_input_size: Optional[int] = None,
+        h_train: int = 1,
+        loss=MAE(),
+        valid_loss=None,
+        max_steps: int = 1,
```
elephaint, line 268: Set `max_steps=0` everywhere so that you can have complete equivalence in the tests

tests/test_models/test_dummy_multivariate.py
```diff
@@ -0,0 +1,197 @@
...
+from neuralforecast.common.enums import TimeSeriesDatasetEnum
+from neuralforecast.utils import PredictionIntervals
+from tests.dummy.dummy_models import DummyMultivariate
+
+
+class TestDummyMultivariate:
+    """Test suite for univariate dummy models to validate horizon predictions functionality."""
+
+    def test_larger_horizon(self, longer_horizon_test):
+        model = DummyMultivariate(
```
elephaint, line 28: Set `max_steps=0` so that you can have perfect equivalence

tests/test_models/test_dummy_multivariate.py
```diff
@@ -0,0 +1,197 @@
+import numpy as np
```
elephaint, line 1: Again this is great!

tests/test_models/test_nlinear.py
```diff
@@ -1,32 +1,82 @@
+import pandas as pd
```
elephaint, line 1: What is special about `NLinear`? I don't think this is a model that is often used so we can also just disable the functionality if this model keeps giving issues, I don't think you have to spend too much time on fixing issues for this model.

tests/helpers/__ini__.py
JQGoh: Fixed in https://github.com/Nixtla/neuralforecast/pull/1368/commits/78dd2e871ac17e642207d35b8e84029a0b17c85c

neuralforecast/common/_base_model.py
JQGoh: Fixed in https://github.com/Nixtla/neuralforecast/pull/1368/commits/9eb95e0860b1ad5c4e8f2450e1ffd98b5593ad65

JQGoh:
> @JQGoh This is awesome!
> 
> I left some comments. Two other things:
> 
> * Where do we test the different loss types for the various dummy models?
> * I'm a bit worried about the introduction of `self.prediction_horizon`; it makes sense from this PR perspective, just need to make sure it always exists and gets reset to default values whenever it is used. So e.g. after calling `.predict`, it should be reset to `self.horizon_backup`

1. For loss types, at first I thought it mainly involve BaseModel logic and tests in DummyUnivariate is sufficient. Having said that, I think it is good to complement the tests for the other models too, so in the future we can also test this easily, see https://github.com/Nixtla/neuralforecast/pull/1368/commits/5d43a46b785c79f4c357e7aa7921daddf1ee1349

2. Agreed that best not to have None. Fix and add tests on this. (PS: also I think this is the most straight forward way to support recurrent prediction, though usually I try not to introduce new attributes)

neuralforecast/common/_base_model.py
JQGoh: Fixed in https://github.com/Nixtla/neuralforecast/pull/1368/commits/3e5c9ea927405427e723c4d438b643af630a8560

neuralforecast/common/_base_model.py
JQGoh: Fixed in https://github.com/Nixtla/neuralforecast/commit/9eb95e0860b1ad5c4e8f2450e1ffd98b5593ad65

neuralforecast/common/_base_model.py
JQGoh: Fixed in https://github.com/Nixtla/neuralforecast/commit/9eb95e0860b1ad5c4e8f2450e1ffd98b5593ad65

tests/dummy/dummy_models.py
JQGoh: By the way, shall we test at least with max_steps=1, to let the models effectively steps forward (that is why I set the learning rate = 0 to reproduce the naive predictions).

I may miss your main point on "complete equivalence" aspect with max_steps=0

tests/test_models/test_nlinear.py
JQGoh: I picked `NLinear` for simplicity. The main purpose of having "test_autonlinear_longer_horizon" test is to check the Auto-kind of model can also support longer horizon predictions due to my implementation in https://github.com/Nixtla/neuralforecast/commit/56c5dd978477a37b10ecf0589573b1edba7e38b2

not planning to deep dive, created an [issue](https://github.com/Nixtla/neuralforecast/issues/1385) for this comment: https://github.com/Nixtla/neuralforecast/pull/1368#discussion_r2345870966

neuralforecast/core.py
JQGoh: We should and this is fixed now

tests/dummy/dummy_models.py
elephaint: Makes sense, I agree to do your version (learning_rate=0.0 and doing 1 step)

I missed the learning_rate=0 part, if that isn't the case I don't think it would be possible to get output equivalence, that's why I mentioned it.

tests/test_models/test_dummy_multivariate.py
elephaint: See other comment, agree with your approach

tests/test_models/test_nlinear.py
elephaint: Ok, thanks!

tests/test_models/test_dummy_univariate.py
```diff
@@ -0,0 +1,325 @@
...
+        else:
+            if isinstance(loss, DISTRIBUTION_LOSSES + MULTIQUANTILE_LOSSES):
+                if isinstance(loss, DISTRIBUTION_LOSSES):
+                    assert model._maybe_get_quantile_idx(quantile) == 1
+                else:
+                    assert model._maybe_get_quantile_idx(quantile) == 0
+            else:
+                model._maybe_get_quantile_idx(quantile) is None
+
+    @pytest.mark.parametrize("loss_type", [IQLoss, HuberIQLoss])
```
elephaint: I'd prefer testing for the following losses:
- A BasePointLoss (MAE) - this should be covered under other scenarios already as it's the default.
- A DistributionLoss  (DistributionLoss(distribution="Normal")
- MQLoss
- IQLoss

And then test every model type (Uni / Multi / RNN)

JQGoh: @elephaint Thanks for this reminder, found a new bug that we fix it. Include the tests for the losses mentioned for different models too

Notable changes since your last review
 * Dummy model did not consider loss.outputsize_multiplier, fixed them in https://github.com/Nixtla/neuralforecast/pull/1368/commits/120776dbcc80d67e65b18b340308d4526cf51384
 * Use of media_idx should be checked with `is not None`, otherwise, media_idx=0 won't be effective. Fixed in https://github.com/Nixtla/neuralforecast/pull/1368/commits/a03e444dcefd8fc3a71af41c76ff2c9d5f902a1e
 
 In addition, I moved the test cases from `test` folder to `tests/test_losses`, otherwise they won't be tested as part of our pytest ci run. See https://github.com/Nixtla/neuralforecast/pull/1368/commits/3ca021e8eee65811e353dab6f94adb4c171b8dff and https://github.com/Nixtla/neuralforecast/pull/1368/commits/6412431facbf15228c87c44ab25c4701a0fe418e

elephaint: That is great, thank you!

PR #1369, opened 2025-08-15
New documentation pipeline

deven367:
This PR completes the transition from `nbdev` to `lazydocs` as the new documentation pipeline. It includes the following changes,

- [x] remove `nbs` folder
- [x] migrate from `numpy` to `google` style docstrings
- [x] have a `pyproject.toml` and `setup.py`, delete `settings.ini`
- [x] remove `nbdev` related workflows
- [x] minor cosmetic changes wrt linting and code formatting
- [x] add test coverage
- [x] change versioning logic
- [x] update docs  
- [x] update contributing md
- [x] add check for broken links and mdx syntax checking

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Unfortunately, available implementations and published research are yet to reali
 
 ## Highlights
 
-* Official `NHITS` implementation, published at AAAI 2023. See [paper](https://ojs.aaai.org/index.php/AAAI/article/view/25854) and [experiments](./experiments/).
+* Official `NHITS` implementation, published at AAAI 2023. See [paper](https://ojs.aaai.org/index.php/AAAI/article/view/25854) and [experiments](https://github.com/Nixtla/neuralforecast/tree/main/experiments).
 * Official `NBEATSx` implementation, published at the International Journal of Forecasting. See [paper](https://www.sciencedirect.com/science/article/pii/S0169207022000413).
 * Unified with`StatsForecast`, `MLForecast`, and `HierarchicalForecast` interface `NeuralForecast().fit(Y_df).predict()`, inputs and outputs.
 * Built-in integrations with `utilsforecast` and `coreforecast` for visualization and data-wrangling efficient methods.


.github/workflows/build-docs.yaml
```diff
@@ -1,66 +1,46 @@
-name: "build-docs"
+name: build-docs
 on:
-  release:
-    types: [released]
+  push:
```
marcopeix: Does that mean that docs are updated whenever there's a push to `main`? I'm not sure that's ideal, because we might add models, and then people see a functionality not yet available.

CONTRIBUTING.md
```diff
@@ -61,35 +67,61 @@ uv pip install -e ".[dev]" --torch-backend cpu # for cpu backend
...
-* Docs are automatically created from the notebooks in the `nbs` folder.
-* In order to modify the documentation:
-    1. Find the relevant notebook.
-    2. Make your changes.
-    3. Run all cells.
-    4. If you are modifying library notebooks (not in `nbs/docs`), clear all outputs using `Edit > Clear All Outputs`.
-    5. Run `nbdev_preview`.
-    6. Clean the notebook metadata using `nbdev_clean`.
+* The docs are automatically generated from the docstrings in the `neuralforecast` folder.
+* To contribute, ensure your docstrings follow the Google style format.
```
marcopeix, line 124: Might be useful to provide a link to what the Google style is, for reference.

.github/workflows/build-docs.yaml
deven367: That's correct! Every time something gets merged in `main`, the docs would get updated

CONTRIBUTING.md
deven367: Done, I've added an example

.github/workflows/build-docs.yaml
marcopeix: I would keep it on release only. We made that change previously, I think it's better that way.

deven367: Done, I've reverted it back to release and PR

PR #1375, opened 2025-08-22
FEAT: Explainability with SHAP in neuralforecast

marcopeix:
Add a method to compute SHAP values for models in neuralforecast. It uses KernelExplainer, a perturbation method, to estimate the contribution of each feature.

The goal of this feature is to work for:
- univariate models
- provide SHAP values for past lags, hist exog, and futr exog
- explain any horizon
- explain any unique_id
- return shap for any univariate model used in `NeuralForecast(models=models)`.

TODO:

- [x] Explain multiple unique_ids
- [x] Explain multiple models
- [x] Additivity holds with `scaler_type`
- [x] Can create plots with shap
- [ ] Unit tests
- [ ] Polars support
- [ ] Option to sum features 
- [x] Smarter wrapper, probably more optimizations to be done to create the samples and make predictions

In a future PR, I would like to include multivariate models (have to see if it's feasible)

### DIFF
--- a/neuralforecast/core.py
+++ b/neuralforecast/core.py
@@ -2,6 +2,7 @@ __all__ = ['NeuralForecast']
 
 
 import pickle
+import shap
 import warnings
 from copy import deepcopy
 from itertools import chain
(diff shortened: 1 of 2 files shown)


elephaint:
Hey, I think this is a great start!

In order for this to work across our other libs I think we need to take a different approach though; both in terms of package that we're using for the explanations and where the explanations are generated. 

Mind if I rewrite so that it naturally works with:
- any model type
- any loss
- any df type
- any accelerator
- different explainers
?

marcopeix:
Delete in favor of #1377

PR #1378, opened 2025-08-31
Fix type of n_windows in cross_validation

Riklia:
As described in the documentation and in the implementation, when test_size is defined, we set n_windows to None. However, the type of `n_windows` is specified as int. This causes a warning on the caller side.

### DIFF
--- a/neuralforecast/core.py
+++ b/neuralforecast/core.py
@@ -1054,7 +1054,7 @@ class NeuralForecast:
         self,
         df: Optional[DataFrame] = None,
         static_df: Optional[DataFrame] = None,
-        n_windows: int = 1,
+        n_windows: Union[int, None] = 1,
         step_size: int = 1,
         val_size: Optional[int] = 0,
         test_size: Optional[int] = None,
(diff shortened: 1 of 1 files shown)


marcopeix:
Hi @Riklia, I want to make a release of neuralforecast soon and would like to include your PR. Can you solve the conflict please? Thanks!

Riklia:
Hi @marcopeix! I rebased, and it should be fine now.

neuralforecast/core.py
```diff
@@ -1337,7 +1337,7 @@ def cross_validation(
         self,
         df: Optional[DataFrame] = None,
         static_df: Optional[DataFrame] = None,
-        n_windows: int = 1,
+        n_windows: Union[int, None] = 1,
```
marcopeix: Can you use `Optional[int]` instead?

neuralforecast/core.py
```diff
@@ -1421,7 +1421,7 @@ def cross_validation(
             test_size = h + step_size * (n_windows - 1)
```
marcopeix, line 1422: Before this line, we need to add an assert for the linter:

`assert n_windows is not None`

neuralforecast/core.py
```diff
@@ -1421,7 +1421,7 @@ def cross_validation(
...
         elif n_windows is None:
             if (test_size - h) % step_size:
```
marcopeix, line 1426: Before the `if` statement, add:

`assert test_size is not None`

neuralforecast/core.py
```diff
@@ -1421,7 +1421,7 @@ def cross_validation(
...
-                raise Exception("`test_size - h` should be module `step_size`")
+                raise Exception("`test_size - h` must be divisible by `step_size`")
             n_windows = int((test_size - h) / step_size) + 1
         else:
             raise Exception("you must define `n_windows` or `test_size` but not both")
```
marcopeix, line 1430: After this line, add these asserts:

```python
assert n_windows is not None
assert test_size is not None
```

neuralforecast/core.py
Riklia: Done. Also added an assertion for `h`. Seems like `self.h` type is `Any`, which is propagated from `models: List[Any]` in the constructor.

Riklia:
Thanks for the detailed instructions! Addressed all the comments.

PR #1380, opened 2025-09-04
Fix: torch version

marcopeix:
Currently a clean install of neuralforecast gives an error when we import models:

```python
from neuralforecast.models import NBEATS, NHITS
from neuralforecast.utils import AirPassengersDF
```

Gives the error: `RuntimeError: operator torchvision::nms does not exist`.

This PR fixes it (tested a clean install in Google Colab)


JQGoh (approved):

valeman:
It appears the same issue still persists in Colab

PR #1063, opened 2024-07-11
[DOCS] Restructure documentation

elephaint:
- This PR restructures the docs in line with TimeGPT's documentation
- Adds missing models to sidebar and doc indices

PR #1381, opened 2025-09-05
Fix broken Open in Colab links

LemuelKL:
- 22 "Open in Colab" embed links in notebooks are broken.
- Caused by commit 4f76f7c11250ec2d6eae28cc7c714cf0b635636a and RP #1063. 
- By the way, the link in nbs/docs/tutorials/08_multivariate_tsmixer.ipynb was pointing to "long horizon with transformers" which is wrong to begin with.
- This fix assumes a link points to the notebook itself, not to another notebook.
- This fix was done programmatically with a Python script. Happy to share if wanted.


marcopeix:
Fixed by #1433

PR #1393, opened 2025-09-24
[FEAT]: any horizon explanation

marcopeix:
Support explanations for horizons that are longer than the training horizon

### DIFF
--- a/neuralforecast/common/_base_model.py
+++ b/neuralforecast/common/_base_model.py
@@ -1394,9 +1394,7 @@ class BaseModel(pl.LightningModule):
         else:
             return y_hat
 
-    def _predict_step_direct(
-        self, batch, batch_idx, recursive=False, trim_window_size=0
-    ):
+    def _predict_step_direct(self, batch, batch_idx, recursive=False, trim_window_size=0):
         temporal_cols = batch["temporal_cols"]
         if recursive:
             # We need to predict recursively, so we use the median quantile if it exists to feed back as insample_y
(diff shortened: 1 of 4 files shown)


neuralforecast/common/_base_model.py
```diff
@@ -1394,36 +1394,121 @@ def _predict_step_recurrent(self, batch, batch_idx):
...
+                all_hist_exog_explanations = []
+                all_stat_exog_explanations = []
+                all_baseline_predictions = []
+            
             for i in range(self.n_predicts):
-                y_hat = self._predict_step_direct(batch, batch_idx, recursive=False)
-
+                # Store the current batch state for explanations
+                if hasattr(self, 'explain') and self.explain:
+                    current_batch = {k: v.clone() if torch.is_tensor(v) else v for k, v in batch.items()}
```
elephaint: Why do we need to clone the batch?

neuralforecast/common/_base_model.py
```diff
@@ -1394,36 +1394,121 @@ def _predict_step_recurrent(self, batch, batch_idx):
...
+                result = self._predict_step_direct(batch, batch_idx, recursive=False)
+                if isinstance(result, tuple):
+                    y_hat = result[0]
+                else:
+                    y_hat = result
+                
                 y_hats.append(y_hat)
+                
+                # Generate explanations for this recursive step if needed
+                if hasattr(self, 'explain') and self.explain:
```
elephaint: I think it looks good, the only thing I'd change is that this function is becoming quite big, so perhaps put the everything below the if-statement here in a separate function?

neuralforecast/common/_base_model.py
```diff
@@ -2026,26 +2129,106 @@ def _explain_batch(
...
+        # Get the horizons to explain
+        all_horizons = self.explainer_config.get("horizons", list(range(self.predict_horizon)))
+        
+        # Filter to only horizons in this recursive step
+        horizons_to_explain = [h for h in all_horizons if step_start <= h < step_end]
+        
+        # Convert to local horizon indices
+        local_horizons = [h - step_start for h in horizons_to_explain]
+        
+        if not local_horizons:
```
elephaint, line 2183: Can you explain this part?

neuralforecast/common/_base_model.py
```diff
@@ -2183,9 +2366,18 @@ def _explain_batch(
                 )
             if add_dim:
                 baseline_predictions = baseline_predictions.unsqueeze(-1)
-            baseline_predictions = baseline_predictions.index_select(1, torch.tensor(horizons, device=baseline_predictions.device))
-            baseline_predictions = baseline_predictions.index_select(2, torch.tensor(series, device=baseline_predictions.device))
-            baseline_predictions = baseline_predictions.index_select(3, torch.tensor(output_index, device=baseline_predictions.device))
+
```
elephaint, line 2395: Why not just use local_horizons in the index select and avoid creating a large new array?

neuralforecast/common/_base_model.py
```diff
@@ -2026,26 +2129,106 @@ def _explain_batch(
...
+        
         # Attribute the input
-        horizons = self.explainer_config.get("horizons", list(range(self.h)))
         series = list(range(self.n_series))
         output_index = self.explainer_config.get(
             "output_index", list(range(y_hat_shape[-1]))
         )
 
         # Start with required inputs
+        insample_y = insample_y.float()
```
elephaint: Why is this necessary? Everything should be float32 already, if it isn't, it feels it should be addressed elsewhere?

elephaint (changes requested):
Looks great @marcopeix, mainly clarifying questions and given how long functions are becoming might be better to split the one in predict up.

neuralforecast/common/_base_model.py
marcopeix: I think we can remove this cloning by reordering the order of functions. Let me try.

neuralforecast/common/_base_model.py
marcopeix: It's to know the index of the prediction horizon for that recursive step. For example, h_train is 6, but horizon is 12, so this allows us know that step 12 is index 5 of the second recursive step. Then, the empty check is to return empty tensors if the `horizons` parameter doesn't explain horizons in that step. Like, `h=12` but `horizons=[0,1]`.

neuralforecast/common/_base_model.py
marcopeix: That was from captum that returned an error of float64 on MPS. I tried to cast everything to float32 in the code, but didn't fix it. It might on Captum's side. The alternative is to move the computation to CPU. I'll remove that, thanks for pointing it out!

neuralforecast/common/_base_model.py
marcopeix: Good point, I changed it.

neuralforecast/common/_base_model.py
elephaint: Thanks!

neuralforecast/common/_base_model.py
```diff
@@ -1393,41 +1393,171 @@ def _predict_step_recurrent(self, batch, batch_idx):
...
+            windows = self._normalization(windows=windows, y_idx=y_idx)
+
+            # Parse windows
+            insample_y, insample_mask, _, _, hist_exog, futr_exog, stat_exog = (
+                self._parse_windows(batch, windows)
+            )
+
+            # Get shape if needed (first iteration of recursive)
+            if first_iteration and y_hat_shape is None:
+                dummy_y_hat = self._predict_step_direct_batch(
```
elephaint: Is this necessary? If you change the order in the for-loop (first create the predictions then the explanations), you can use y_hat_shape from that?

neuralforecast/common/_base_model.py
```diff
@@ -1393,41 +1393,171 @@ def _predict_step_recurrent(self, batch, batch_idx):
...
+                            all_stat_exog_explanations.append(stat_exog)
+                        if baseline is not None:
+                            all_baseline_predictions.append(baseline)
+                
+                # Temporarily disable explanations for the recursive call
+                if explain_state:
+                    self.explain = False
+                
+                # Make predictions for this step
                 y_hat = self._predict_step_direct(batch, batch_idx, recursive=False)
```
elephaint: Do this first, then the explanations, then you have y_hat_shape and don't need the dummy.

neuralforecast/common/_base_model.py
```diff
@@ -1529,6 +1681,7 @@ def _predict_step_direct(
                 y_hats.append(y_hat)
 
             y_hat = torch.cat(y_hats, dim=0)
+            
             if hasattr(self, "explain") and self.explain:
```
elephaint: Shouldn't this also be "explain_state"?

neuralforecast/common/_base_model.py
```diff
@@ -1449,7 +1579,29 @@ def _predict_step_direct(
 
             y_hat = torch.cat(y_hats, dim=1)
             self.test_size = total_test_size
+            
+            # Return with concatenated explanations if explaining
+            if explain_state:
```
elephaint, line 1568: Can you change it everywhere to this?

neuralforecast/common/_base_model.py
```diff
@@ -2026,26 +2176,106 @@ def _explain_batch(
...
+        
         # Attribute the input
-        horizons = self.explainer_config.get("horizons", list(range(self.h)))
         series = list(range(self.n_series))
         output_index = self.explainer_config.get(
             "output_index", list(range(y_hat_shape[-1]))
         )
 
         # Start with required inputs
+        insample_y = insample_y.float()
```
elephaint: What was the cause of this being necessary? Does MPS cast stuff to float64?

neuralforecast/common/_base_model.py
```diff
@@ -1393,41 +1393,171 @@ def _predict_step_recurrent(self, batch, batch_idx):
             )        
         else:
             return y_hat
-
-    def _predict_step_direct(
-        self, batch, batch_idx, recursive=False, trim_window_size=0
+        
+    def _compute_explanations_for_step(
```
elephaint, line 1398: 👍

elephaint (changes requested):
Nice @marcopeix, few questions / changes

neuralforecast/common/_base_model.py
marcopeix: Yes, there was a problem with MPS and float64 that forced me to use CPU instead. But I meant to remove all of those `.float()`. Thanks for pointing it out!

neuralforecast/common/_base_model.py
marcopeix: Good point!

PR #1399, opened 2025-10-16
[docs] lazydocs to mkdocstrings

deven367:
This PR migrates from lazydocs to mkdocstrings

- [x] Update Makefile
- [x] Update build-docs workflow
- [x] Add parser
- [x] Add docs
- [x] Correctly handles math equations

### DIFF
--- a/docs/tsdataset.html.md
+++ b/docs/tsdataset.html.md
@@ -10,12 +10,20 @@ title: PyTorch Dataset/Loader
 ::: neuralforecast.tsdataset.TimeSeriesLoader
 
 ::: neuralforecast.tsdataset.BaseTimeSeriesDataset
+    options:
+      show_if_no_docstring: false
 
 ::: neuralforecast.tsdataset.LocalFilesTimeSeriesDataset
+    options:
+      show_if_no_docstring: false
 
 ::: neuralforecast.tsdataset.TimeSeriesDataset
+    options:
+      show_if_no_docstring: false
 
 ::: neuralforecast.tsdataset.TimeSeriesDataModule
+    options:
+      show_if_no_docstring: false
 
 ### Example
 


.github/workflows/build-docs.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
           user_email: 41898282+github-actions[bot]@users.noreply.github.com
 
       - name: Trigger mintlify workflow (Push to main or Pull Request)
-        if: (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'pull_request' || (github.event_name == 'workflow_dispatch' && github.event.inputs.environment == 'staging')
+        if: github.event_name == 'release' || (github.event_name == 'workflow_dispatch' && github.event.inputs.environment == 'production')
```
deven367: This condition doesn't look right, it needs to be for `pull_request` and `staging`

nasaul: Thanks Deven, already changed it

Makefile
```diff
@@ -13,14 +13,14 @@ load_docs_scripts:
...
-	lazydocs .neuralforecast --no-watermark
-	python docs/to_mdx.py
+	python docs/to_mdx.py docs
 
 examples_docs:
 	mkdir -p nbs/_extensions
 	cp -r docs-scripts/mintlify/ nbs/_extensions/mintlify
 	python docs-scripts/update-quarto.py
 	quarto render nbs/docs --output-dir ../docs/mintlify/
+	find docs/mintlify/ -name "*.mdx" ! -name "*.html.mdx" -type f -exec sh -c 'mv "$$1" "$${1%.mdx}.html.mdx"' _ {} \;
```
deven367: Is this only for the jupyter notebooks? when we convert them to `mdx`?

docs/models.autoformer.html.md
```diff
@@ -0,0 +1,130 @@
+---
+output-file: models.autoformer.html
+title: Autoformer
+---
```
deven367, line 6: should we also add `description` here and for the other models as well?

nasaul: There were some models that didn't have any description and other that had. I decided to unify them all and left it as an introuduction to all the models. We can discuss if it's better with the description or just the introduction.

deven367: It's mostly from an SEO perspective where `ahrefs` complains if the description tag is missing for the documents

Makefile
nasaul: The notebooks convert into `mdx` when we add to `_quarto.yml` the project type:

``` yaml
project:
  type: mintlify
```

Then I guess the command should be: 

``` bash
find docs/mintlify/docs/ -name "*.mdx" ! -name "*.html.mdx" -type f -exec sh -c 'mv "$$1" "$${1%.mdx}.html.mdx"' _ {} \;
```

Do you agree?

docs/models.autoformer.html.md
nasaul: Ok, then I should update the introduction parts into `description`

deven367: We can get some more information from @khuyentran1401  as well about this, she might have also additional insights about it

Makefile
deven367: yes, makes sense!

deven367:
This is amazing work @nasaul! It's time to merge 🎉

PR #1404, opened 2025-10-29
[FEAT] Add support for historical and static exogenous in TimeXer

marcopeix:
TimeXer now supports all types of exogenous features.

Example usage:

```python
Y_train_df = AirPassengersPanel[AirPassengersPanel['ds'] < AirPassengersPanel['ds'].values[-12]].reset_index(drop=True)
Y_test_df = AirPassengersPanel[AirPassengersPanel['ds'] >= AirPassengersPanel['ds'].values[-12]].reset_index(drop=True)
futr_df = Y_test_df.drop(columns=["y", "y_[lag12]"])

models = [
    TimeXer(
        h=12,
        input_size=24,
        n_series=2,
        hist_exog_list=["y_[lag12]"],
        futr_exog_list=["trend"],
        stat_exog_list=['airline1'],
        max_steps=20,
        scaler_type="robust",
    ),
]

nf = NeuralForecast(
    models=models, 
    freq="ME", 
)

nf.fit(
    df=Y_train_df,
    static_df=AirPassengersStatic
)
fcsts = nf.predict(futr_df=futr_df)
```

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2447,9 +2447,8 @@ dependencies = [
     { name = "numpy", version = "2.3.2", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.11'" },
     { name = "optuna" },
     { name = "pandas" },
-    { name = "pydantic" },
     { name = "pytorch-lightning" },
-    { name = "ray", extra = ["tune"] },
+    { name = "ray", extra = ["train", "tune"] },
     { name = "torch" },
     { name = "utilsforecast" },
 ]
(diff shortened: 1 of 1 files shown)


JQGoh (commented):
A couple of questions and requests:
 * Suggest to add a minimal example of test using exogenous variables for  in `tests/test_models/test_timexer.py`. Consider examples like 1) X with static exog, 2) X without static exog  
 * I notice that you include `pydantic` dependency in this PR, is it because you ran into Ray-related failure and this helps to resolve it? If that is the case, do you still remember what kind of Ray error?
 
 Otherwise, the PR LGTM

marcopeix:
@JQGoh 

1. Sure, I can add a test, good idea!
2. Right, I added pydantic because of ray issues. The error message was 
```
ImportError: `ray.train.v2` requires the pydantic package, which is missing. Run the following command to fix this: `pip install pydantic`
```
So I added the dependency to fix the issue.

pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
     "numpy>=1.21.6",
     "pandas>=1.3.5",
     "torch>=2.4.0",
+    "pydantic>=2.0.0",
```
elephaint: The cleaner way to do this is add the dependency at ray. See [here](https://github.com/ray-project/ray/issues/58280)

pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
...
     "pytorch-lightning>=2.0.0",
     "ray[tune]>=2.2.0",
```
elephaint: "ray[tune,train]>=2.2.0"

marcopeix:
For the tests, I agree we should have them, but I noticed that none of the models that support exogenous features have tests for exog features. I can make a separate PR to add a global test for models that support exog feat, something similar to the current `test_model`. And keep this PR only for TimeXer. Wdyt?

JQGoh:
yes, I also notice that. Sounds good to me

> For the tests, I agree we should have them, but I noticed that none of the models that support exogenous features have tests for exog features. I can make a separate PR to add a global test for models that support exog feat, something similar to the current `test_model`. And keep this PR only for TimeXer. Wdyt?

elephaint (approved):
LGTM, nice!

non-blocking - you can use "check_model" from .common._model_checks with "losses" to test models against each loss function, this typically uncovers bugs (I didn't run it on this PR, just "fyi" in case you didn't know)

PR #1406, opened 2025-11-09
[FEAT] Support normalization for static exogenous features

JuditHalperin:
This PR adds normalization for static exogenous features via `local_static_scaler_type` in `NeuralForecast`, extending the existing normalization framework that already supports `local_scaler_type` for temporal features.

### **Motivation**

Static features can vary widely in magnitude (e.g., a static feature such as building volume, where different buildings might have values of 500, 1,000, and 20,000). Without normalization, these differences can:
* Slow down training convergence
* Cause numerical instability when combined with temporal inputs
* Bias learning toward large-magnitude features

Unlike temporal inputs, statics are constant per entity and do not vary by window, so only local scaling is meaningful.

### **Implementation**

Added new argument:

```python
NeuralForecast(..., local_static_scaler_type="standard")
```

### DIFF
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -42,8 +42,8 @@ jobs:
       - name: Build Docs
         run: make all_docs
 
-      - name: Deploy (Push to main or Pull Request)
-        if: github.event_name == 'pull_request' || (github.event_name == 'workflow_dispatch' && github.event.inputs.environment == 'staging')
+      - name: Deploy (Push to main or Pull Request from same repo)
+        if: (github.event_name == 'push' && github.ref == 'refs/heads/main') || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository) || (github.event_name == 'workflow_dispatch' && github.event.inputs.environment == 'staging')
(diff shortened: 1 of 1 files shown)


neuralforecast/core.py
```diff
@@ -234,6 +235,8 @@ def __init__(
             freq (str or int): Frequency of the data. Must be a valid pandas or polars offset alias, or an integer.
             local_scaler_type (str, optional): Scaler to apply per-serie to all features before fitting, which is inverted after predicting.
```
marcopeix: Instead of "all features", could we change it to "temporal features" please?

Yudit42: @marcopeix I've updated the docs, thanks!

Yudit42:
Hi @marcopeix, all checks passed, but the `build-docs` check failed due to a permission error. Is there anything I can do?

marcopeix:
> Hi @marcopeix, all checks passed, but the `build-docs` check failed due to a permission error. Is there anything I can do?

@JuditHalperin , no nothing you can do on your end, this is entirely on us! We will get this sorted out and merged!

marcopeix (approved):

PR #1410, opened 2025-11-15
Align UR2CUTE integration with upstream release (auto-threshold, masked losses, tests)

FH-Prevail:
## Summary
- mirror the latest standalone UR2CUTE estimator: wire BaseModel knobs, enable padding/sanity overrides, and add auto-threshold bookkeeping + masked BCE/regression losses
- tighten inference by using the learned threshold and exposing a `_return_components` flag so training/validation can consume both CNN outputs safely
- skip validation when `val_size == 0` to keep Lightning happy on short datasets and add pytest smoke coverage (fit/predict + auto-threshold)

## Testing
- python -m pytest tests/test_models/test_ur2cute.py -q --cov-fail-under=0


FH-Prevail:
Hi together,
What are the steps I should do at this moment?
for detail about the method , I already published it in a journal here:
https://www.mendel-journal.org/index.php/mendel/article/view/413

elephaint:
Hi, thanks for your contribution!

I've read the paper; whilst forecasting for intermittent products poses an interesting problem, I think the work lacks the quality needed to include this in Neuralforecast at this moment:
- I feel the architecture is insufficiently detailed and motivated. Why a CNN when currently MLPs and Transformers show better performance on time series data? Why use both the classifier and regressor? 
- The data used in the experiment is undisclosed; the provided github doesn't open source data or experiment, making it impossible to reproduce the experiments; 
- Odd choices were made; why use a complex combination of MAE and MSE when HuberLoss exists? 
- No comparison was made against established neural network models;
- Key experimental details are missing; for example, what is the forecasting horizon?
- The experiments insufficiently validate the claims made in the paper; For example, the use of the loss mechanism isn't ablated, so there's no evidence that the classifier+regressor setup provides value to the model. Similarly for the (odd) combination of MAE and MSE.
- The conclusions says "_In the case of UR2CUTE, we combined two CNNs – one for frequency and one for magnitude of demand. Our results indicate a significant improvement and show how combining multiple methods can improve forecasting accuracy for intermittent demand._" None of the experiments validate this conclusion. The experiments don't show how the combination adds value to the prediction (showing a single picture in Figure 4 is insufficient).

To fix the significant issues I'd encourage the author to take inspiration from strong neural network architecture papers, for example:
- [DeepAR](https://www.sciencedirect.com/science/article/pii/S0169207019301888)
- [NHITS](https://ojs.aaai.org/index.php/AAAI/article/view/25854) ([Arxiv](https://arxiv.org/abs/2201.12886))
- [PatchTST](https://github.com/yuqinie98/PatchTST)

A potential path to inclusion in NeuralForecast would be to demonstrate the method's performance up to a higher standard, for which the provided papers are pointers. However, this is not a guarantee that we will include it.

As regards the implementation, the modification of the validation- and training_step is unnecessary based on my understanding of the method; using a custom loss function would be sufficient for this method. 

Please don't take this message as a discouragement. I'm sharing this advice to support you and help you move forward with the best possible foundation.

Issue #1357, opened 2025-07-09
[BUG] Memory Scalability Issue with Large Number of Exogenous Variables and Large-Scale Dataloader


strakehyr:
### What happened + What you expected to happen

When using the NeuralForecast.fit() method with the large-scale dataloader (by passing a list of Parquet files to the df argument), the GPU memory usage appears to scale inefficiently with the number of exogenous variables. I could run these specs before (used neuralforecast when it was version 1.7) and trian a model with them. With all the updates since then, I had to try the Large dataloading Method and still the problem only arises when exogenous variables are enabled. Beware that I can train TSMixer on these many Exog variables on other libraries, so theres definitely some memory handling issue. 

Training a standard model (e.g., TSMixerx with default hyperparameters) on a large dataset with many time steps but no exogenous variables works as expected, successfully streaming data with a low memory footprint. However, adding a large number of exogenous variables (~80-90) to hist_exog_list and futr_exog_list causes a CUDA out of memory error on a high-VRAM GPU (48 GB), even when the model's architecture and batch_size remain unchanged.

This suggests that the windowing process for exogenous variables may not be as memory-efficient as it is for the target variable y when using the file-based dataloader, potentially loading the entire feature set for a series into memory at once.

### Versions / Dependencies

neuralforecast = 3.0.2
pytorch_lightning = 2.5.2

### Reproduction script
```

import numpy as np
import pandas as pd
import tempfile
import os
from neuralforecast import NeuralForecast
from neuralforecast.models import TSMixerx

# 1. Define dataset parameters
n_series = 10
n_timesteps = 800_000
n_exog = 90 # The high number of features that causes the issue
h = 24      # Forecast horizon
input_size = 96 # Lookback window

# 2. Create a large, synthetic DataFrame
print("Creating synthetic DataFrame...")
ids = [f'series_{i}' for i in range(n_series)]
df_list = []
for i, unique_id in enumerate(ids):
    # Create timestamps
    start_date = pd.to_datetime('2020-01-01')
    timestamps = pd.date_range(start=start_date, periods=n_timesteps, freq='15min')
    
    # Create data
    y = np.random.randn(n_timesteps)
    
    # Create DataFrame for this series
    series_df = pd.DataFrame({
        'ds': timestamps,
        'unique_id': unique_id,
        'y': y
    })
    
    # Add exogenous features
    for j in range(n_exog):
        series_df[f'exog_{j}'] = np.random.randn(n_timesteps)
        
[...]

PR #1421, opened 2025-12-11
[FIX] Models train with many exogenous features when using files

marcopeix:
As described in #1357, if we currently pass a list of files with a high number of exogenous features, we get OOM errors. 

To fix it, I propose to truncate large datasets and "rotate" between different chunks of data such that we load reasonable amounts of data into memory and the model can see all the data given enough training steps.

Original bug was reproduced in Colab using the script below. This PR fixes the bug.

```python
# 1. Define dataset parameters
n_series = 10
n_timesteps = 100_000
n_exog = 90 # The high number of features that causes the issue
h = 24      # Forecast horizon
input_size = 96 # Lookback window

# 2. Create a large, synthetic DataFrame
print("Creating synthetic DataFrame...")
ids = [f'series_{i}' for i in range(n_series)]
df_list = []
for i, unique_id in enumerate(ids):
    # Create timestamps
    start_date = pd.to_datetime('2020-01-01')
    timestamps = pd.date_range(start=start_date, periods=n_timesteps, freq='15min')

    # Create data
    y = np.random.randn(n_timesteps)

    # Create DataFrame for this series
    series_df = pd.DataFrame({
        'ds': timestamps,
        'unique_id': unique_id,
        'y': y
    })

    # Add exogenous features
    for j in range(n_exog):
        series_df[f'exog_{j}'] = np.random.randn(n_timesteps)

    df_list.append(series_df)

Y_df = pd.concat(df_list).reset_index(drop=True)

# Ensure data types are memory-efficient
numeric_cols = Y_df.select_dtypes(include=np.number).columns
Y_df[numeric_cols] = Y_df[numeric_cols].astype(np.float32)
print("DataFrame created. Shape:", Y_df.shape)


# 3. Save as partitioned Parquet files for large-scale loading
print("Saving data to partitioned Parquet files...")
tmpdir = tempfile.TemporaryDirectory()
Y_df.to_parquet(tmpdir.name, partition_cols=['unique_id'], index=False)

files_list = [
    f"{tmpdir.name}/{d}"
    for d in os.listdir(tmpdir.name)
    if os.path.isdir(f"{tmpdir.name}/{d}") and d.startswith('unique_id=')
]
static_df = pd.DataFrame({'unique_id': ids})


# 4. Define model and NeuralForecast object
exog_list = [f'exog_{j}' for j in range(n_exog)]

# Using standard, non-intensive hyperparameters
model = TSMixerx(
    h=h,
    input_size=input_size,
    n_series=n_series,
    hist_exog_list=exog_list, # This is the key part that causes the issue
    # futr_exog_list=exog_list, # Can also be used here
    revin=True,
    scaler_type='standard',
    batch_size=32, # Using a reasonable batch size,
    max_steps=10,
)

nf = NeuralForecast(models=[model], freq='15min')

# 5. Attempt to fit the model
# This step is expected to cause a CUDA OOM error due to memory handling of exogenous variables
import traceback

try:
    print("Attempting to fit the model...")
    nf.fit(
        df=files_list,
        static_df=static_df,
        val_size=h * 10
    )
except Exception as e:
    print(f"\nTraining failed. Error: {e}")
    traceback.print_exc()

# Clean up
tmpdir.cleanup()
```

### DIFF
--- a/neuralforecast/tsdataset.py
+++ b/neuralforecast/tsdataset.py
@@ -500,10 +500,6 @@ class LocalFilesTimeSeriesDataset(BaseTimeSeriesDataset):
         temporal_cols = self.temporal_cols.copy()
         path = self.files_ds[idx]
         
-        # Get and increment access count for rotation
-        access_count = self._access_counts[idx]
-        self._access_counts[idx] += 1
-        
         # Get total row groups for this file
         total_row_groups = self.row_groups_per_file[idx]
         
@@ -518,17 +514,27 @@ class LocalFilesTimeSeriesDataset(BaseTimeSeriesDataset):
         groups_needed = max(1, (self.max_size + rows_per_group - 1) // rows_per_group)
         groups_needed = min(groups_needed, total_row_groups)
         
-        # Determine which row groups to read based on rotation
-        rotation_cycle = access_count // self.rotation_frequency
-        start_group = (rotation_cycle * groups_needed) % total_row_groups
-        
-        # Handle wraparound if needed
-        if start_group + groups_needed <= total_row_groups:
-            groups_to_read = list(range(start_group, start_group + groups_needed))
+        # For small datasets, load all groups in order to ensure deterministic training
+        # For large datasets, use rotation to cycle through different portions
+        if self.max_size <= 10_000:
+            # Load all row groups sequentially
+            groups_to_read = list(range(groups_needed))
         else:
-            # Wrap around to beginning
-            groups_to_read = list(range(start_group, total_row_groups)) + \
-                            list(range(0, groups_needed - (total_row_groups - start_group)))
+            # Get and increment access count for rotation
+            access_count = self._access_counts[idx]
+            self._access_counts[idx] += 1
+            
+            # Determine which row groups to read based on rotation
+            rotation_cycle = access_count // self.rotation_frequency
+            start_group = (rotation_cycle * groups_needed) % total_row_groups
+            
+            # Handle wraparound if needed
+            if start_group + groups_needed <= total_row_groups:
+                groups_to_read = list(range(start_group, start_group + groups_needed))
+            else:
+                # Wrap around to beginning
+                groups_to_read = list(range(start_group, total_row_groups)) + \
+                                list(range(0, groups_needed - (total_row_groups - start_group)))
             
         # Read the row groups
         if Path(path).is_dir():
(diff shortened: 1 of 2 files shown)


neuralforecast/tsdataset.py
```diff
@@ -456,43 +457,149 @@ def __init__(
...
+            self.row_groups_per_file.append(total_groups)
+            self.total_rows_per_file.append(total_rows)
 
     def __getitem__(self, idx):
         if not isinstance(idx, int):
             raise ValueError(f"idx must be int, got {type(idx)}")
 
+        import pyarrow.parquet as pq
+        import pyarrow as pa
+        from pathlib import Path
```
elephaint, line 498: This one should be at the top, it's a generic import

elephaint (commented):
Whilst I think the solution could fix the issue, I'm a bit worried about what is stated in the issue:

>  I could run these specs before (used neuralforecast when it was version 1.7) and trian a model with them

This suggests we made a change that causes the OOM. Shouldn't we debug/understand what change is causing the OOM? Because previously this should not be a problem.

If we look at the toy data:
```
n_series * n_timesteps * n_exog = 10 * 100_000 * 90 = 90M datapoints
4 bytes per float 32 = 90 * 4 = ~343 MB
```

So the data shouldn't be an issue.

The maximum window created should be no bigger than:

```
[batch_size, h + L, n_features, n_series] = [32, 96 + 24, 90, 10]
32 * (96 + 24) * 90 * 10 = 3.45M
3.45M x 4 bytes per float32 = ~13.8MB
```
So the window shouldn't be an issue either.

If I'd had to guess, maybe we are creating windows for the entire dataset on the GPU?

[This line](https://github.com/Nixtla/neuralforecast/blob/c1490d15a82450f8b0fc6ddeeee9f6def43401c5/neuralforecast/common/_base_model.py#L727) maybe gets "realized", i.e., normally unfolding is just a view (not a copy), but perhaps it's being realized.

At some point I put the window sampling into a separate [subfunction](https://github.com/Nixtla/neuralforecast/blob/c1490d15a82450f8b0fc6ddeeee9f6def43401c5/neuralforecast/common/_base_model.py#L887). 

Could it be that by splitting these two, the windows are "realized" when they leave the function "create_windows", effectively overloading the GPU?

I guess my question is - what's causing the issue? I feel we fix it but we don't understand the root cause, as the data size suggests this shouldn't be a problem in our training script.

neuralforecast/tsdataset.py
gitvold: I now often get the OutOfMemoryError: CUDA out of memory error with my RTX 3060 (12 GB VRAM). It seems this issue depends mainly on input_size (and possibly dataset size), and not on batch size or model parameters. The actual memory consumption is 5–20x higher than I expected.

marcopeix:
PR #1429 fixes the issue and this PR is not relevant anymore.

Issue #1297, opened 2025-03-21
Mac MINI M4 2024 Forecast on GPU Memory Issue. On CPU is fine


Mogiaro:
### What happened + What you expected to happen

Hello i'm using latest version 3.0.0 on Python 3.9 or 3.10 on a brand new .venv

What i get is that when i'm training a model for a "small" dataframe reading about 6MB of csv (attached), on CPU (accelerator="cpu") it goes fine and even with 27 days concat it goes fine. When i remove the accelerator line the error i got prevent me to fit the model since it's triggering the GPU enabled. This is so strange in my opinion. I can attach the .csv data and the notebook i'm using. 

[message.txt](https://github.com/user-attachments/files/19392621/message.txt)

Review of the Error: 
RuntimeError: Invalid buffer size: 17.62 GB 

NOTE: 17.62 is just a case, if i use a different structure or change something it could be also 42 GB and it feels really strange since Neural Networks should work on so little data in my few experience...

Can you help in some way? Thanks!

### Versions / Dependencies

Python 3.9/10 on MacMini latest OS. 
Libs are the default doing 

pip install neuralforecast==3.0.0 

### Reproduction script

[bug-report-memory.ipynb.zip](https://github.com/user-attachments/files/19392604/bug-report-memory.ipynb.zip)

[bug_report_data.csv](https://github.com/user-attachments/files/19392792/bug_report_data.csv)

### Issue Severity

High: It blocks me from completing my task.

marcopeix:
Hello! Can you try reducing the batch size? That usually solves out-of-memory errors. For example, you could do:
```python
nhits = NHITS(
    h=horizon, 
    input_size=input_size, 
    batch_size=32,
    windows_batch_size=32
)
```
Let me know if this works!

Mogiaro:
No, changing batch size is not make it work. I tried 32, 16, 8

Mogiaro:
With n_blocks=2 it does change the num of params but the error of buffer size still remains. I could try another model, but i believe the buffer size problem for GPU will still remain

elephaint:
> With n_blocks=2 it does change the num of params but the error of buffer size still remains. I could try another model, but i believe the buffer size problem for GPU will still remain

What difference did the other suggestions make?

I was also thinking, try to include only data that the model also consumes. So, if the model doesn't use the exogenous variables in your data, drop them from the input dataframe. I think we currently lack a prefilter for unused data.

Mogiaro:
[...]

PR #1429, opened 2026-01-05
[FIX] Apply windows batch size

marcopeix:
Applies `windows_batch_size` before permuting to avoid materializing many windows into memory and running into OOM issues. 

With this change we fix issues #1357 and #1297.

This PR also replaces #1421; that PR will not be needed and this is a simpler fix.

### DIFF
--- a/neuralforecast/common/_base_model.py
+++ b/neuralforecast/common/_base_model.py
@@ -1310,7 +1310,7 @@ class BaseModel(pl.LightningModule):
         for i in range(n_batches):
             # Create and normalize windows [Ws, L+H, C]
             w_idxs = torch.arange(
-                i * windows_batch_size, min((i + 1) * windows_batch_size, n_windows, device=windows_temporal.device)
+                i * windows_batch_size, min((i + 1) * windows_batch_size, n_windows), device=windows_temporal.device
             )
             windows = self._sample_windows(
                 windows_temporal=windows_temporal,


neuralforecast/common/_base_model.py
```diff
@@ -722,6 +722,12 @@ def _create_windows(self, batch, step):
                 dimension=-1, size=window_size, step=self.step_size
             )
 
+            # Sample windows before permute
```
elephaint: Shouldn't we delete the existing sampling method then?

neuralforecast/common/_base_model.py
```diff
@@ -722,6 +722,12 @@ def _create_windows(self, batch, step):
...
+            # windows shape: [n_series, C, Ws, window_size]
+            if self.windows_batch_size is not None and windows.shape[2] > self.windows_batch_size:
```
elephaint: We can't sample before the creation of the windows because the available conditions establish which windows are eligible to sample from

elephaint (changes requested):
I don't think this works; the sampling should be after we've established the eligible windows (otherwise we are sampling highly inefficiently).

Where are the windows materialized? What line of code? That's where the issue is.

elephaint:
Edit: doesn't work - still same error 😞

neuralforecast/common/_base_model.py
```diff
@@ -742,37 +742,58 @@ def _create_windows(self, batch, step):
...
-                    outsample_condition, axis=(1, -1)
-                )  # Sum over time & series dimension
-                final_condition = (outsample_condition >= min_outsample_points) & (
-                    insample_condition >= min_insample_points
-                )
+            
+            if self.MULTIVARIATE:
+                # For multivariate: windows shape is [Ws, L + h, C, n_series]
+                insample_condition = windows[:, : self.input_size, available_idx, :]
+                insample_condition = torch.sum(insample_condition, axis=(1, 2))
```
elephaint: The summing axis should be over the time and series dimension

PR #1434, opened 2026-01-13
[CHORE] Bump version, drop Python 3.9, add Python 3.13

marcopeix:
Prepare to release v.3.1.3

Adds support for Python 3.13
Drops support for Python 3.9

Note that Ray is still in beta for Python 3.13 on Windows, that's why we exclude that Python version in the tests for Windows.


elephaint (approved):
LGTM!

We could also consider adding 3.14 already as it's [stable](https://devguide.python.org/versions/), but it's not really necessary.

PR #1461, opened 2026-02-15
fix: correct min_samples formula in conformal prediction for step_size

### LINKED ISSUE #1352: Error message related to step_size in conformal prediction methods

1. The bug
When we use a conformal prediction method, it returns an error message if we set the `n_windows` too high. For example, for `horizon = 52 `and `n_windows = 3`, we have

`ValueError: Minimum required samples in each serie for the prediction intervals settings are: 157, shortest serie has: 156. Please reduce the number of windows, horizon or remove those series`

2. Expected behavior
For a `step_size` equal to 1, we would expect this to work, and not get an error message. In fact, the minimum time series length should be 55, not 157.

3. Useful information
In the method [_conformity_scores](https://github.com/Nixtla/neuralforecast/blob/d1348cd46eef039a0d4727e82278aaa8ea832642/neuralforecast/core.py#L1661), we have : 
 
```
min_size = ufp.counts_by_id(df, id_col)["counts"].min()
min_samples = self.h * self.prediction_intervals.n_windows + 1
if min_size < min_samples:
raise ValueError(
                "Minimum required samples in each serie for the prediction intervals "
                f"settings are: {min_samples}, shortest serie has: {min_size}. "
                "Please reduce the number of windows, horizon or remove those series."
            )

self._add_level = True
cv_results = self.cross_validation(
df=df,
       static_df=static_df,
       n_windows=self.prediction_intervals.n_windows,
       id_col=id_col,
       time_col=time_col,
       target_col=target_col,
)

```

Thus, the code returns an error message if the time series is shorter than `horizon*n_windows+1`. This assumes that the step_size used is equal to the horizon, by default.
However, it then uses the [cross_validation](https://github.com/Nixtla/neuralforecast/blob/d1348cd46eef039a0d4727e82278aaa8ea832642/neuralforecast/core.py#L1089) function, defined as follows:

```
    def cross_validation(
        self,
        df: Optional[DataFrame] = None,
        static_df: Optional[DataFrame] = None,
        n_windows: int = 1,
        step_size: int = 1,
        val_size: Optional[int] = 0,
        test_size: Optional[int] = None,
        use_init_models: bool = False,
        verbose: bool = False,
        refit: Union[bool, int] = False,
        id_col: str = "unique_id",
        time_col: str = "ds",
        target_col: str = "y",
        prediction_intervals: Optional[PredictionIntervals] = None,
        level: Optional[List[Union[int, float]]] = None,
        quantiles: Optional[List[float]] = None,
        **data_kwargs,
    )
```

By default, the step_size is equal to 1 (and we have no control over this parameter when we call the PredictionIntervals() class). So there should be no restrictions of this type when using conformal prediction methods.
Maybe we should have this restriction : 
`min_samples = self.h + step_size + 1`

Instead of : 
`min_samples = self.h * self.prediction_intervals.n_windows + 1`

### Versions / Dependencies

`neuralforecast == 3.0.0`

### Reproduction script

```
import numpy as np
import pandas as pd

from neuralforecast.utils import PredictionIntervals
from neuralforecast.losses.pytorch import DistributionLoss, MAE
from datasetsforecast.m4 import M4

df,X_df,S_df=M4().load(directory='data', group='Weekly', cache=True)
df["ds"] = pd.to_numeric(df["ds"])

horizon = 12
input_size = 24
prediction_intervals = PredictionIntervals(method="conformal_error", n_windows = 8)

models = [
 NHITS(
     h=horizon,
     input_size=input_size,
     max_steps=100,
    loss=MAE(),
     scaler_type="robust",
  )
]

nf = NeuralForecast(models=models, freq=1)
nf.fit(df, prediction_intervals=prediction_intervals)

expected_df = nf.make_future_dataframe(df)
preds = nf.predict(futr_df=expected_df, level=[90])

```

Returns : 
`ValueError: Minimum required samples in each serie for the prediction intervals settings are: 97, shortest serie has: 93. Please reduce the number of windows, horizon or remove those series.`

However, the step_size is set to 1 in the cross_validation method so we do not need 97 samples in each series, but less.

Moreover, I think it would be particularly useful to be able to choose the step_size in conformal prediction methods!

### Issue Severity

W057:
## Summary

- Fixes the `min_samples` check in `_conformity_scores()` which incorrectly assumed `step_size == h`, causing `ValueError` for valid series lengths
- Adds `step_size` parameter to `PredictionIntervals` so users can control the step size used in conformal cross-validation
- The corrected formula `h + step_size * n_windows` (suggested by @elephaint) matches the actual `cross_validation()` behavior

## Changes

**`neuralforecast/utils.py`**
- Added `step_size` parameter to `PredictionIntervals.__init__()` (default `1`, backward compatible)
- Added validation (`step_size >= 1`)

**`neuralforecast/core.py`**
- Changed formula from `h * n_windows + 1` to `h + step_size * n_windows`
- Passes `step_size` to `cross_validation()` call

## Test plan

- [x] Verified the reproduction script from #1352 no longer raises `ValueError` (series length 93, h=12, n_windows=8)
- [x] Verified custom `step_size` works (step_size=6, n_windows=3)
- [x] Default `step_size=1` preserves backward compatibility

Closes #1352

### DIFF
--- a/neuralforecast/core.py
+++ b/neuralforecast/core.py
@@ -1715,6 +1715,15 @@ class NeuralForecast:
             save_dataset (bool): Whether to save dataset or not.
             overwrite (bool): Whether to overwrite files or not. 
         """
+        # In distributed training (DDP), only rank 0 should save
+        try:
+            import torch.distributed as dist
+
+            if dist.is_initialized() and dist.get_rank() != 0:
+                return
+        except (ImportError, RuntimeError):
+            pass
+
         # Standarize path without '/'
         if path[-1] == "/":
             path = path[:-1]
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1040,6 +1040,46 @@ def test_save_load_no_dataset(setup_airplane_data):
     forecasts2 = fcst2.predict(df=AirPassengersPanel_train, futr_df=AirPassengersPanel_test)
     np.testing.assert_allclose(forecasts1["DilatedRNN"], forecasts2["DilatedRNN"])
 
+
+def test_save_skips_nonzero_ddp_rank(monkeypatch, tmp_path):
+    """Only rank 0 should write artifacts when DDP is initialized."""
+    dist = pytest.importorskip("torch.distributed")
+    monkeypatch.setattr(dist, "is_initialized", lambda: True)
+    monkeypatch.setattr(dist, "get_rank", lambda: 1)
+
+    fcst = NeuralForecast(
+        models=[DilatedRNN(h=12, input_size=-1, encoder_hidden_size=5, max_steps=1)],
+        freq="M",
+    )
+    save_path = tmp_path / "ddp_rank_1_save"
+    fcst.save(path=str(save_path), model_index=None, overwrite=True, save_dataset=False)
+
+    assert not save_path.exists()
+
+
+def test_save_runs_on_rank0_ddp(monkeypatch, tmp_path, setup_airplane_data):
+    """Rank-0 DDP process must save all artifacts normally."""
+    dist = pytest.importorskip("torch.distributed")
+
+    AirPassengersPanel_train, _ = setup_airplane_data
+    fcst = NeuralForecast(
+        models=[DilatedRNN(h=12, input_size=-1, encoder_hidden_size=5, max_steps=1)],
+        freq="M",
+    )
+    fcst.fit(AirPassengersPanel_train)
+
+    save_path = tmp_path / "ddp_rank_0_save"
+    with monkeypatch.context() as m:
+        m.setattr(dist, "is_initialized", lambda: True)
+        m.setattr(dist, "get_rank", lambda: 0)
+        fcst.save(
+            path=str(save_path), model_index=None, overwrite=False, save_dataset=False
+        )
+
+    assert save_path.exists()
+    assert (save_path / "alias_to_model.pkl").exists()
+    assert (save_path / "configuration.pkl").exists()
+
 # test `enable_checkpointing=True` should generate chkpt
 def test_enable_checkpointing(setup_airplane_data):
     AirPassengersPanel_train, _ = setup_airplane_data


marcopeix:
- branch needs to be updated
- proper tests need to be included to show it works and safeguards are in place

W057:
During validation of this PR branch, we found an unrelated issue while running the broad test suite (excluding distributed tests): 244 passed, 17 failed, 44 deselected. The persistent failures are in test_neural_forecast_scaling and the explainability block (CPUAccelerator devices=-1 when accelerator='cpu'). This PR remains scoped to conformal step_size/min_samples behavior and tests. Could someone else confirm reproduction? If confirmed, we can open a separate issue to track it.

tests/test_models/test_dummy_univariate.py
```diff
@@ -140,6 +140,22 @@ def test_conformal_prediction(self, longer_horizon_test):
         assert "DummyUnivariate-lo-80" in forecasts.columns
         assert "DummyUnivariate-hi-80" in forecasts.columns
 
+    def test_conformal_prediction_min_samples_uses_step_size(self, longer_horizon_test):
```
marcopeix, line 143: Can you add a positive path to the test? Like this:

```python
def test_conformal_prediction_min_samples_uses_step_size(self, longer_horizon_test):
      short_train_df = (
          longer_horizon_test.train_df.groupby(TimeSeriesDatasetEnum.UniqueId)
          .tail(10)
          .reset_index(drop=True)
      )
      model = DummyUnivariate(h=longer_horizon_test.h, input_size=4)
      nf = NeuralForecast(models=[model], freq="ME")

      # Positive case: step_size=1 with 10 samples should NOT raise (min_samples=7)
      nf.fit(
          df=short_train_df,
          prediction_intervals=PredictionIntervals(n_windows=3, step_size=1),
      )

      # Negative case: step_size=3 requires 11, 10 is not enough
      with pytest.raises(
          ValueError, match=r"settings are: 11, shortest serie has: 10\."
      ):
          nf.fit(
              df=short_train_df,
              prediction_intervals=PredictionIntervals(n_windows=3, step_size=3),
          )
```

W057:
Addressed the latest review request.

Updated `test_conformal_prediction_min_samples_uses_step_size` to include both:
- Positive path (`step_size=1`) with enough samples to fit successfully.
- Negative path (`step_size=3`) asserting the expected min-samples error.

Validation run:
- `uv run pytest --no-cov tests/test_models/test_dummy_univariate.py::TestDummyUnivariate::test_conformal_prediction_min_samples_uses_step_size tests/test_utils.py::test_prediction_intervals_step_size_validation -q`
- Result: `3 passed`

Latest commit: `a7bceb8a`.

PR #1464, opened 2026-02-20
[FEAT] Explainability for multivariate models

marcopeix:
Adds explainability for multivariate models, which is currently disabled.

To consider:
- [x] Doesn't work with `stat_exog` yet. We need a batch dimension for that. Need to check if it's fairly simple to patch.
- [x] Update documentation

### DIFF
--- a/neuralforecast/core.py
+++ b/neuralforecast/core.py
@@ -1172,22 +1172,17 @@ class NeuralForecast:
                 f"You can always set outputs=None to default to [0] (first output)."
             )
 
-        # Determine minimum series across all models to explain
-        min_series = min(getattr(model, 'n_series', 1) for model in models_to_explain)
+        # Validate series. All multivariate models share the same n_series;
+        # univariate models always use [0] and ignore this parameter.
+        mv_models = [m for m in models_to_explain if m.MULTIVARIATE]
+        n_series = mv_models[0].n_series if mv_models else 1
 
-        # Validate series
         if series is None:
-            series = list(range(min_series))
-        elif not series or any(s < 0 or s >= min_series for s in series):
-            per_model = ", ".join(
-                f"{m.hparams.alias if hasattr(m.hparams, 'alias') and m.hparams.alias else m.__class__.__name__}="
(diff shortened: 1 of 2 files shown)


marcopeix:
> Nice!
> 
> I think just one thing:
> 
> The scaling on this is pretty horrible: O(horizons x series x outputs x n_steps). I think we need to add a "series" argument to "explain" (this was also in the original PR, we ditched it because multivariate wasn't implemented yet), which by default should be None, so that only features of a sub selection can be computed.
> 
> (and after adding this, it needs docstring, addition to the example, etc)

Ah, forgot about that parameter! You're right, we should add it back

neuralforecast/core.py
```diff
@@ -1169,7 +1171,25 @@ def explain(
                 f"You must set valid output indices for all models, which is the minimum number of ouputs amongst all models. "
                 f"You can always set outputs=None to default to [0] (first output)."
             )
-        
+
+        # Determine minimum series across all models to explain
+        min_series = min(getattr(model, 'n_series', 1) for model in models_to_explain)
```
elephaint: Trying to think what this means - it is either 1 or the total amount of series, right?

Can we even initiate a NF class with e.g. multivariate(10) and multivariate(20)? That would be weird.

neuralforecast/core.py
```diff
@@ -1169,7 +1171,25 @@ def explain(
...
+        # Validate series
+        if series is None:
+            series = list(range(min_series))
+        elif not series or any(s < 0 or s >= min_series for s in series):
+            per_model = ", ".join(
+                f"{m.hparams.alias if hasattr(m.hparams, 'alias') and m.hparams.alias else m.__class__.__name__}="
+                f"{getattr(m, 'n_series', 1)}"
+                for m in models_to_explain
+            )
+            raise ValueError(
```
elephaint, line 1183: Same for this error mode - not sure what situation it covers; it's either 1 or n_series? So min_series is either 1 or n_series?

elephaint (commented):
Nice!! LGTM - one question before we can merge imho

PR #1473, opened 2026-03-05
[FEAT] Use dates for train/val/test split

marcopeix:
Add the option to use dates instead of number of time steps to perform the train/val/test split in cross-validation.

### DIFF
--- a/neuralforecast/common/_modules.py
+++ b/neuralforecast/common/_modules.py
@@ -41,63 +41,72 @@ ACTIVATIONS = [
 
 
 class MLP(nn.Module):  
-    """Multi-Layer Perceptron for time series forecasting.  
+    """Multi-Layer Perceptron for time series forecasting.
 
-    A feedforward neural network with configurable depth and width. The network  
-    consists of an input layer, multiple hidden layers with activation functions  
-    and dropout, and an output layer. All hidden layers have the same dimensionality.  
+    A feedforward neural network with configurable depth and width. The network
+    consists of an input layer, multiple hidden layers with activation functions
+    and dropout, and an output layer. All hidden layers have the same dimensionality.
 
-    Args:  
-        in_features (int): Dimension of input features.  
-        out_features (int): Dimension of output features.  
-        activation (str): Activation function name. Must be one of the supported  
-            activations in ACTIVATIONS list (e.g., 'ReLU', 'Tanh', 'GELU', 'ELU').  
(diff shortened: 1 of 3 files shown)


neuralforecast/core.py
```diff
@@ -1396,6 +1423,8 @@ def cross_validation(
         step_size: int = 1,
         val_size: Optional[int] = 0,
         test_size: Optional[int] = None,
+        validation_cutoff: Optional[Any] = None,
```
elephaint, line 1426: Allowing Any here is a validation nightmare, especially with dates that can be timestamp, pandas offsets, strings....

elephaint (changes requested):
I get the intuitive part.... but:

- Input validation is tough:
   - What if df doesn't have timestamps
   - What if validation date is later than test date
   - Different timestamp formats
- Kinda breaks API consistency across all the repos (we'd have to add it in cv endpoints sf / mf / nixtla too) - doesn't mean we shouldn't implement it, but just noting.
- val_size and test_size are used quite extensively throughout the library; if we support this in the cv endpoint, would make sense to make it consistent and support it in all endpoints where test_size and val_size are provided as an argument.

If kept very consistent throughout NF and strong input validation, we could use it, but it remains hard....

E.g. "2011-03-02", how does it know what date that is if the format isn't supplied? So, it should then check against the format of the df I guess?

marcopeix:
> I get the intuitive part 😄 .... but:
> 
>     * Input validation is tough:
>       
>       * What if df doesn't have timestamps
>       * What if validation date is later than test date
>       * Different timestamp formats
> 
>     * Kinda breaks API consistency across all the repos (we'd have to add it in cv endpoints sf / mf / nixtla too) - doesn't mean we shouldn't implement it, but just noting.
> 
>     * val_size and test_size are used quite extensively throughout the library; if we support this in the cv endpoint, would make sense to make it consistent and support it in all endpoints where test_size and val_size are provided as an argument.
> 
> 
> If kept very consistent throughout NF and strong input validation, we could use it, but it remains hard....
> 
> E.g. "2011-03-02", how does it know what date that is if the format isn't supplied? So, it should then check against the format of the df I guess?

I agree, and I'm not sure if I like this feature either haha 😅. I knew validation would be hard, and I still find it fragile. To be honest, it hasn't been a wildly requested feature. Let's store it aside for now, and see if it's something the community really wants.

PR #1478, opened 2026-03-16
[FEAT] Cache trainer for faster inference

marcopeix:
A small change that caches the trainer so it's not recreated every time. This speeds up consecutive inference almost 2x. Useful when the same model is used for inference multiple times and during cross-validation (when `refit=False`). Below is the result of a small profiling benchmark:

### CPU

| Config | Cold (ms) | Warm (ms) | Speedup |
|---|---|---|---|
| Baseline (no cache) | 109.7 ± 14.5 | 103.8 ± 14.9 | 1.00x |
| Trainer cache | 104.1 ± 21.7 | 53.3 ± 8.3 | 1.95x |

### GPU (Tesla T4)

| Config | Cold (ms) | Warm (ms) | Speedup |
|---|---|---|---|
| Baseline (no cache) | 137.2 ± 41.3 | 126.7 ± 34.8 | 1.00x |
| Trainer cache | 111.7 ± 5.4 | 66.7 ± 9.6 | 1.90x |

> **Cold** = first `predict()` call (includes Trainer init overhead).
> **Warm** = mean of 15 subsequent calls — the number that matters for repeated inference.
> **Speedup** = baseline warm / config warm.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -2300,8 +2300,14 @@ def test_trainer_caching(setup_airplane_data):
     nf.predict()
     assert model._pred_trainer is trainer_1, "Trainer should be reused on repeated predict calls"
 
-    # Changing pred_trainer_kwargs invalidates the cache
+    # Adding a new key invalidates the cache
     cached_kwargs = model._pred_trainer_kwargs.copy()
     model._pred_trainer_kwargs = {**cached_kwargs, "__dummy__": True}
     nf.predict()
-    assert model._pred_trainer is not trainer_1, "Trainer should be replaced when kwargs change"
+    trainer_2 = model._pred_trainer
+    assert trainer_2 is not trainer_1, "Trainer should be replaced when a kwarg is added"
+
+    # Changing an existing value also invalidates the cache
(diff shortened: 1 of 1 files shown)


neuralforecast/common/_base_model.py
```diff
@@ -2070,7 +2070,13 @@ def predict(
                 'baseline_predictions': baseline_predictions
             }
         else:
-            trainer = pl.Trainer(**pred_trainer_kwargs)
+            if (
+                not hasattr(self, "_pred_trainer")
+                or self._pred_trainer_kwargs != pred_trainer_kwargs
```
elephaint, line 2075: Just checking - does this ensure a one-to-one identical key-value mapping of the pred_trainer_kwargs?

marcopeix: Yes

marcopeix:
> Great idea! One question re. changing the kwargs (the test tests adding a kwarg, does it also fail when changing a kwarg to a different value?)

Good point, I added a case where a kwarg is changed which invalidates the cache

elephaint (approved):

PR #1480, opened 2026-03-17
[FIX] Make spark extra dep installable via pip

marcopeix:
Add `spark` to `[project.optional-dependencies]` so it's installable via `pip`.


marcopeix:
not required since #1481

PR #1481, opened 2026-03-17
[FIX] Spark to optional-dependencies

nasaul:
- [x] Adds Spark to optional-dependencies
- [x] Updates every reference from `--group spark` to `--extra spark`
- [x] Automatically updates the version for `mkdocstring-parser` so we don't need to update the `uv.lock` with each correction of `mkdocstring-parser`


JQGoh (approved):
LGTM

PR #1482, opened 2026-03-18
[FEAT] Sample paths for recurrent models

marcopeix:
Adds the `return_samples` argument to return sample paths for recurrent models. Now, models like DeepAR or any recurrent model (LSTM with `recurrent=True`) can return sample paths when used with `DistributionLoss`.


elephaint:
I think this doesn't work, as on each new step, the mean is used of the previous generated distributional step to generate the next sample paths. That is not correct sample path generation (it should use each N path generated and roll that forward).

marcopeix:
Same is achieved in #1483, but the latter generalizes to all models.

PR #1483, opened 2026-03-20
[FEAT] Simulation paths

elephaint:
### Summary

Adds `NeuralForecast.simulate()`, a method for generating temporally correlated sample paths from fitted models. Unlike `predict(level=...)` which produces independent per-step intervals, `simulate()` generates full joint trajectories that preserve temporal correlation, enabling scenario analysis.

### New API

```python
nf.simulate(
    n_paths=100,          # number of sample paths
    seed=42,              # reproducibility
    method="gaussian_copula",  # or "schaake_shuffle"
)
# Returns DataFrame: [unique_id, ds, sample_id, model_1, model_2, ...]
```

### Simulation Methods

| Method | Description |
|--------|-------------|
| `gaussian_copula` (default) | AR(1) Gaussian copula with Toeplitz correlation and IQF marginals ([Baron et al., 2025](https://arxiv.org/abs/2510.02224)) |
| `schaake_shuffle` | Nonparametric — independent marginal samples reordered by historical rank templates ([Clark et al., 2004](https://journals.ametsoc.org/view/journals/hydr/5/1/1525-7541_2004_005_0243_tssamf_2_0_co_2.xml)) |

We don't (yet) support the (baseline) "roll-forward" autoregressive simulation. This is for a number of reasons:
- The roll-forward is very slow (requires H predict calls instead of 1). To do it efficiently, requires substantial rewrite in base_model.py
- The roll-forward suffers from noise / variance explosion (I think in Baron et al. they call it snowball effect)

### Documentation
**`nbs/docs/tutorials/simulation.ipynb`** — Tutorial demonstrating different loss types, decision-making example, and a side-by-side comparison of Gaussian copula vs. Schaake shuffle.

### Open issues (I suggest to leave open for now)
- `h > self.h` (simulating with horizon larger than train horizon) is not yet supported. This is intentional, as it requires substantial more work (and testing) to feed back the generated sample paths autoregressively for both recurrent and direct models. See also why the roll-forward approach hasn't been implemented.
- explainability also not supported for simulated paths.

### DIFF
--- a/neuralforecast/utils.py
+++ b/neuralforecast/utils.py
@@ -742,12 +742,12 @@ def interp_2d(x, xp, fp):
     corresponding row of ``fp`` as knot values.
 
     Args:
-        x: (H, n_paths) query points in (0, 1).
+        x: (N, n_paths) query points in (0, 1).
         xp: (Q,) shared sorted knot positions.
-        fp: (H, Q) per-row knot values.
+        fp: (N, Q) per-row knot values.
 
     Returns:
-        result: (H, n_paths) interpolated values.
+        result: (N, n_paths) interpolated values.
     """
     Q = xp.shape[0]
     idx = torch.searchsorted(xp.expand(x.shape[0], -1), x).clamp(1, Q - 1)
@@ -770,7 +770,8 @@ def gaussian_copula_sample(
 ):
     """Gaussian copula with AR(1) Toeplitz correlation and IQF marginals.
 
-    All computation in torch tensors.
+    All computation in torch tensors. Cholesky decompositions and sampling
+    are batched across series for efficiency.
 
     Args:
         quantile_positions: (Q,) sorted quantile levels in (0, 1), tensor.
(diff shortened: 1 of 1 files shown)


neuralforecast/common/_base_model.py
```diff
@@ -2098,6 +2098,122 @@ def predict(
 
         return fcsts
 
+    def simulate(
+        self,
+        dataset,
+        n_paths=500,
```
marcopeix, line 2104: Isn't this a bit too much? Maybe 100 is a more reasonable default?

tests/test_simulate.py
```diff
@@ -0,0 +1,814 @@
...
+import pytest
+import torch
+
+
+def _get_model_cols(df):
+    """Get model columns (everything except unique_id, ds, sample_id)."""
+    return [c for c in df.columns if c not in ("unique_id", "ds", "sample_id")]
+
+
+class TestSimulation:
```
marcopeix, line 14: Would be good to have a test when we have `futr_exog`.

marcopeix (changes requested):
Very nice, and the tutorial is great too! Some minor comments.

neuralforecast/common/_base_model.py
elephaint: Good point, yeah

PR #1525, opened 2026-05-19
[FEAT] Can pass `run_config` with Ray

### LINKED ISSUE #1374: `neuralforecast` should allow to passthrough tune `runConfig`
### Description

Currently, the `neuralforecast` uses default values for `runConfig()` in the tuning function.
Unfortunately, the default of Ray is bad as the logging is set to write outside the working directory, in user's home. I don't believe there is any good reason to store these logging there, but regardless of my opinion, users should have control over this when using `AutoNHITS`, etc.

- https://github.com/Nixtla/neuralforecast/blob/fbb012c856094ff9ab88d007bd04042c01c555a6/neuralforecast/common/_base_auto.py#L246-L257

`neuralforecast` should at least allow to change the storage path and, better, set a default that is localised within the working directory.

- https://docs.ray.io/en/latest/tune/tutorials/tune-output.html#how-to-configure-logging-in-tune
- https://docs.ray.io/en/latest/tune/api/doc/ray.tune.RunConfig.html#ray.tune.RunConfig

### Use case

_No response_

marcopeix:
Auto models with `backend=ray` can now take a `run_config` to pass different options, such as storage path.

### DIFF
--- a/neuralforecast/common/_base_auto.py
+++ b/neuralforecast/common/_base_auto.py
@@ -64,6 +64,11 @@ class BaseAuto(pl.LightningModule):
         callbacks (list of callable): List of functions to call during the optimization process.
             ray reference: https://docs.ray.io/en/latest/tune/tutorials/tune-metrics.html
             optuna reference: https://optuna.readthedocs.io/en/stable
+        run_config (ray.air.RunConfig, optional): Ray Tune `RunConfig` forwarded to `tune.Tuner`.
+            Use this to customize Ray Tune output.
+            When provided, the user-supplied `RunConfig` is used as-is, so `callbacks`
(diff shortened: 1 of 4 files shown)


neuralforecast/common/_base_auto.py
```diff
@@ -244,9 +251,14 @@ def _tune_model(
             else None
         )
 
+        if self.run_config is not None:
+            run_config = self.run_config
```
nasaul, line 260: NIT: Raising a warning could make this case more explicit.

``` python
if self.callbacks is not None:
          warnings.warn(
              "`callbacks` is ignored when `run_config` is provided; "
              "set callbacks on the RunConfig instead.",
          )
      run_config = self.run_config
```

nasaul (approved):
LGTM

PR #1527, opened 2026-05-20
[FEAT] Add study kwargs for Optuna

### LINKED ISSUE #1231: [<Library component: Models|Core|etc...>] Customize some parameters for Study object (Optuna backend)
### Description

Hi.

It would be nice if it were possible to customize the Study object (Optuna backend) for all Auto models.

Sometimes the optimization process breaks with a 'RuntimeError' (or any other error). This happens because occasionally, depending on the parameters sampled for the model, a DataFrame full of null values is generated, and applying the validation metric causes an error. It would be possible to prevent the optimization process from stopping if the Study could be customized using the 'study.optimize(..., catch=(RuntimeError,))' attribute. The optimization process would simply ignore this error and proceed to the next one.

This could be achievied if we could pass some "study_kwargs" in the Auto* models creation, I think.

For example:

AutoDeepAR(
    h=24,
    config=config_deepar,
    search_alg=optuna.samplers.TPESampler(seed=1989),
    backend='optuna',
    num_samples=5,
    study_kwargs={'catch' : '(RuntimeError,)'}
)

I hope I made myself clear.

Thanks for your attention.

### Use case

_No response_

marcopeix:
Add `study_kwargs` for Optuna backend to catch errors or pass other parameters to `study.optimize()`

### DIFF
--- a/neuralforecast/auto.py
+++ b/neuralforecast/auto.py
@@ -86,6 +86,7 @@ class AutoRNN(BaseAuto):
         backend="ray",
         callbacks=None,
         run_config=None,
+        study_kwargs=None,
     ):
         # Define search space, input/output sizes
         if config is None:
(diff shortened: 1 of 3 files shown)


nasaul (approved):
LGTM

PR #1529, opened 2026-05-21
[FIX] Reset`EarlyStopping` state in cross-validation when `refit=True`

### LINKED ISSUE #1124: How refit on train+val set?
How should I refit the model on the whole dataset containing both training and validation sets?

It seems that both `nf.fit` and `nf.cross_validation` don't support this feature, but AutoModels have `refit_with_val`. Could I set `refit_with_val` on non-auto models?

marcopeix:
Currently, each refit window inherits the `EarlyStopping` state. Since each window has a different validation set, the best score is stale and patience counter is not reset.

Now, `EarlyStopping` callback is reset for each consecutive fit.

### DIFF
--- a/neuralforecast/common/_base_model.py
+++ b/neuralforecast/common/_base_model.py
@@ -8,7 +8,7 @@ import warnings
 from contextlib import contextmanager
 from copy import deepcopy
 from dataclasses import dataclass
-from typing import Dict, List, Union
+from typing import Dict, List, Optional, Union
 
 import fsspec
 import numpy as np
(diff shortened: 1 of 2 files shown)


neuralforecast/common/_base_model.py
```diff
@@ -604,6 +604,18 @@ def _fit(
...
+        # Reset here so every fit starts with a fresh early-stopping counter.
+        for cb in self.trainer_kwargs.get("callbacks") or []:
+            if isinstance(cb, EarlyStopping):
+                cb.wait_count = 0
+                cb.stopped_epoch = 0
+                cb.best_score = (
+                    torch.tensor(torch.inf)
+                    if cb.mode == "min"
+                    else torch.tensor(-torch.inf)
+                )
```
gee-senbong: Any reason why not fixing this at the cross validation function? Would it be cleaner if we just construct a new `EarlyStopping` instance. By doing so, we don't need to handle these internal details manually. 

Unfortunately `pytorch_lightning` does not have `reset` method (https://docs.pytorch.org/ignite/generated/ignite.handlers.early_stopping.EarlyStopping.html#ignite.handlers.early_stopping.EarlyStopping.reset).

marcopeix: Don't want to do it just for cross-validation but for consecutive call to `fit` too. Which is why I don't fix it at the cross-validation function. But, I could rebuilt it from the supplied kwargs by the user. I think that's cleaner.

gee-senbong (approved):

PR #1531, opened 2026-05-22
[FEAT] Add ray scheduler and refactor optuna and ray kwargs

### LINKED ISSUE #1080: Specify Ray Scheduler / BOHB Support

marcopeix:
Add Ray Tune scheduler passthrough and refactor BaseAuto backend-specific kwargs into typed config objects

`BaseAuto` now forwards a user-supplied scheduler to `tune.TuneConfig`, unlocking non-FIFO schedulers like ASHA, HyperBand, and BOHB.

**Refactor: RayOptions / OptunaOptions**
Adding scheduler made it clear that the growing list of backend-specific top-level kwargs (run_config, scheduler for Ray, study_kwargs, create_study_kwargs for Optuna) was about to churn 35 Auto* subclass signatures every time a new option is added.
  
Replaced them with two typed dataclasses:

`AutoMLP(..., ray_options=RayOptions(scheduler=ASHAScheduler(), run_config=...))`
`AutoMLP(..., optuna_options=OptunaOptions(study_kwargs={"catch": (RuntimeError,)}))`

- Future backend options become a one-line dataclass field instead of a 35-subclass signature change.
- The four separate "wrong backend" warning blocks collapse into one helper.
- The four legacy kwargs remain accepted for one release with a DeprecationWarning pointing at the new API. Passing both a legacy kwarg and its dataclass equivalent raises TypeError.

**Deprecation timeline**
Legacy kwargs (run_config, scheduler, study_kwargs, create_study_kwargs) are scheduled for removal in v3.2.0.


neuralforecast/common/_base_auto.py
```diff
@@ -64,22 +150,20 @@ class BaseAuto(pl.LightningModule):
...
-            See https://optuna.readthedocs.io/en/stable/reference/generated/optuna.create_study.html.
+        ray_options (RayOptions, optional): Container for Ray-only options. See
+            `RayOptions` for the supported fields (`run_config`, `scheduler`).
+            Only used with `backend='ray'`.
+        optuna_options (OptunaOptions, optional): Container for Optuna-only options.
+            See `OptunaOptions` for the supported fields (`study_kwargs`,
+            `create_study_kwargs`). Only used with `backend='optuna'`.
+        run_config: Deprecated, will be removed in v3.2.0. Pass
+            `ray_options=RayOptions(run_config=...)` instead.
+        scheduler: Deprecated, will be removed in v3.2.0. Pass
```
nasaul: `scheduler` was defined in this PR, right? I don't see the reason to document for deprecation.

neuralforecast/common/_base_auto.py
```diff
@@ -99,7 +183,10 @@ def __init__(
         alias=None,
         backend="ray",
         callbacks=None,
+        ray_options=None,
+        optuna_options=None,
         run_config=None,
+        scheduler=None,
```
nasaul: Why use `scheduler` when it's going to get deprecated?

neuralforecast/common/_base_auto.py
nasaul: This call could become problematic between the old version and the new one, suggest using: `ignore=["run_config", "scheduler", "study_kwargs", "create_study_kwargs"]`

neuralforecast/common/_base_auto.py
```diff
@@ -127,18 +214,28 @@ def __init__(
...
-            )
+
+        ray_options = ray_options if ray_options is not None else RayOptions()
+        optuna_options = (
+            optuna_options if optuna_options is not None else OptunaOptions()
+        )
+        _coalesce_legacy_option(
+            "run_config", run_config, ray_options, "run_config", "ray_options"
+        )
+        _coalesce_legacy_option(
```
nasaul: Do we need this one?

neuralforecast/common/_base_auto.py
```diff
@@ -127,18 +214,28 @@ def __init__(
...
+            "study_kwargs", study_kwargs, optuna_options, "study_kwargs", "optuna_options"
+        )
+        _coalesce_legacy_option(
+            "create_study_kwargs",
+            create_study_kwargs,
+            optuna_options,
+            "create_study_kwargs",
+            "optuna_options",
+        )
+        _warn_unused_options(backend, ray_options, optuna_options)
```
nasaul, line 220: Fires twice when the legacy kwarg targets the wrong backend (e.g. `backend='ray', study_kwargs={...}`): a DeprecationWarning from `_coalesce_legacy_option` pointing at
 `OptunaOptions(study_kwargs=...)`, then a UserWarning here saying that target is ignored. Suggest passing backend into
  `_coalesce_legacy_option` and skipping the deprecation when the `kwarg` belongs to the inactive backend.

nasaul (commented):
The addition of the scheduler is quite a nice idea. However this PR introduces some weird stuff for the new introduced parameters.

neuralforecast/common/_base_auto.py
marcopeix: Nice catch, solved

neuralforecast/common/_base_auto.py
```diff
@@ -92,23 +150,24 @@ def __init__(
...
     ):
         super(BaseAuto, self).__init__()
         with warnings.catch_warnings(record=False):
             warnings.filterwarnings("ignore")
             # the following line issues a warning about the loss attribute being saved
             # but we do want to save it
-            self.save_hyperparameters()  # Allows instantiation from a checkpoint from class
+            # Ignore deprecated kwargs so they aren't persisted into checkpoints
+            # and break loads after they're removed in v3.2.0.
+            self.save_hyperparameters(ignore=["cpus", "gpus"])
```
nasaul: Do we want to save the full `ray_options` and `optuna_options`? If not, then I would argue that it would be better something like: 

``` python
ignore=["cpus", "gpus", "ray_options", "optuna_options"]
```

nasaul (approved):
Great job! Just think if we want to apply the comment that I left. Otherwise good to go

PR #1539, opened 2026-05-28
[FIX] Prevent a second optimization study when using conformal predictions

### LINKED ISSUE #1232: AutoNHITS with prediction_intervals creates second optimization study

When using AutoNHITS with prediction_intervals, I noticed it creates a new optimization study and reruns the entire hyperparameter search process a second time. Example:

```python
model = AutoNHITS(h=24, num_samples=2, backend="optuna")
nf = NeuralForecast(models=[model], freq="H")
nf.fit(df=data, prediction_intervals=PredictionIntervals(n_windows=2))
```

The logs show "A new study created in memory" after the initial training completes, and it runs another complete optimization cycle.

Is this intended behavior? I would have expected it to run one hyperparameter optimization to find the best configuration, and then use the best model to generate prediction intervals.

### Versions / Dependencies

Python 3.11, Neural Forecast 1.77

### Reproduction script

import numpy as np
import pandas as pd
from neuralforecast.auto import AutoNHITS
from neuralforecast.utils import PredictionIntervals

# Create mock data
np.random.seed(42)
n_samples = 1000
dates = pd.date_range("2020-01-01", periods=n_samples, freq="h")
data = pd.DataFrame(
    {"ds": dates, "unique_id": "A", "y": np.random.normal(0, 1, n_samples)}
)

# Configure model
model = AutoNHITS(
    h=24,  # Forecast horizon
    num_samples=2,  # Number of trials
    backend="optuna",
)

# Create NeuralForecast wrapper
nf = NeuralForecast(models=[model], freq="H")

# Fit model - this will show sampling running twice
print("Starting model fit...")
nf.fit(df=data, prediction_intervals=PredictionIntervals(n_windows=2))
print("Finished model fit")

### Issue Severity

Medium: It is a significant difficulty but I can work around it.

marcopeix:
Currently, Auto models with conformal predictions will fire a second optimization. This PR prevents it by reusing the optimal config to compute the prediction intervals.

### DIFF
--- a/neuralforecast/core.py
+++ b/neuralforecast/core.py
@@ -651,6 +651,14 @@ class NeuralForecast:
                     f"{train_size} timestamp(s) available for training after removing val_size."
                 )
 
+        # `_conformity_scores` (above) already ran the Auto* search and left the
+        # results on the current models. Capture them before any reset so the search
+        # can be reused for the final fit instead of rerunning on the full dataset.
+        auto_search_results = [
+            getattr(model, "results", None) if isinstance(model, BaseAuto) else None
+            for model in self.models
+        ]
+
         # Recover initial model if use_init_models
         if use_init_models:
             self._reset_models()
@@ -660,6 +668,10 @@ class NeuralForecast:
         reuse_auto_search = self._cs_df is not None
         for i, model in enumerate(self.models):
             if reuse_auto_search and isinstance(model, BaseAuto):
+                # `_reset_models` swaps in fresh clones without results; restore the
+                # captured search results so the reuse guard in BaseAuto.fit passes.
+                if auto_search_results[i] is not None:
+                    model.results = auto_search_results[i]
                 model._reuse_search = True
             try:
                 self.models[i] = model.fit(
--- a/neuralforecast/core.py
+++ b/neuralforecast/core.py
@@ -655,10 +655,19 @@ class NeuralForecast:
         if use_init_models:
             self._reset_models()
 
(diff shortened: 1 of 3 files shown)


tests/test_core.py
```diff
@@ -2126,6 +2126,53 @@ def test_neuralforecast_conformal_prediction(setup_airplane_data, setup_airplane
     assert "uid" in preds.columns
     assert any(col.startswith(model.__name__) for col in preds.columns)
 
+def test_auto_model_prediction_intervals_runs_search_once(setup_airplane_data, monkeypatch):
```
elephaint: This test only exercises the optuna backend, but ray is the default (backend="ray") and the reuse path branches on self.backend for best_config extraction (_base_auto.py:587). Should we test on ray too?

neuralforecast/core.py
```diff
@@ -655,10 +655,19 @@ def fit(
         if use_init_models:
             self._reset_models()
 
+        # When `_conformity_scores` has already run the Auto* search, mark the
```
elephaint, line 666: When use_init_models=True, _conformity_scores runs the search on the current models, but _reset_models() then swaps in new clones with no results, so the reuse guard fails and the search re-runs.

elephaint (changes requested):
Nice, two comments

neuralforecast/core.py
marcopeix: Good catch, fixed it for this path

elephaint (approved):
LGTM!

PR #1540, opened 2026-05-28
[FIX] Raise error when missing params in auto config

### LINKED ISSUE #929: [Auto] Auto* models can not be used with any `config` as a subset of `default_config`
### Description

Few models (e.g. `TFT`) takes required positional arguments (e.g. `input_size`). If user passes a `config` to `AutoTFT` that does not have this key, training fails with following as the root cause:

> TypeError: TFT.__init__() missing 1 required positional argument: 'input_size'

Here's a reproducible example:

```python
import pandas
from neuralforecast.auto import AutoTFT
from ray import tune

from sktime.datasets import load_longley
from sktime.split import temporal_train_test_split

y, X = load_longley()
y_train, y_test, X_train, X_test = temporal_train_test_split(y, X, test_size=4)

algorithm = AutoTFT(
    4,
    config={"max_steps": tune.choice([5, 10, 15]), "random_seed": tune.choice([0])},
    backend="ray",
)
model = NeuralForecast([algorithm], "A-DEC")

train_data = {
    "unique_id": 1,
    "ds": y_train.index.to_timestamp(freq="A-DEC").to_numpy(),
    "y": y_train.to_numpy(),
}
for column in X.columns:
    train_data[column] = X_train[column].to_numpy()

train_dataset = pandas.DataFrame(data=train_data)

model.fit(df=train_dataset)
# RuntimeError: No best trial found for the given metric: loss. This means that no trial has reported this metric, or all values reported for this metric are NaN. To not ignore NaN values, you can set the `filter_nan_and_inf` arg to False.
```

It will be helpful to document mandatory keys to be passed in `config` in the documentation of `Auto*` models.

If it is possible to allow users to pass few arugments to the underlying model directly without tuning, for `__init__` (where tuning is not necessary) or to `fit` or `predict` (for example configurations for PyTorch Lighning's trainer), it will be very useful.

### Link

_No response_

marcopeix:
Currently, if a config for an auto model has a missing required parameter, it raises only an error when it reaches the hyperopt functions. Now, we raise an error immediately at construction.

### DIFF
--- a/neuralforecast/common/_base_auto.py
+++ b/neuralforecast/common/_base_auto.py
@@ -1,6 +1,7 @@
 __all__ = ['BaseAuto', 'RayOptions', 'OptunaOptions']
 
 
+import inspect
 import warnings
 from copy import deepcopy
 from dataclasses import dataclass, fields, replace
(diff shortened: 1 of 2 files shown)


elephaint (approved):
LGTM!

PR #1544, opened 2026-06-01
[FEAT] Remove cpus and gpus in favor of ray_options

### LINKED ISSUE #1535: Deprecate duplicate parameters in Auto models
Deprecate `cpus` and `gpus` in favor or `ray_options`

marcopeix:
We now pass `cpus` and `gpus` in `ray_options`.


neuralforecast/auto.py
elephaint, line 1: Doesn't this break API?

elephaint (commented):
Isn't this a breaking change?

neuralforecast/auto.py
marcopeix: Yes, we introduced a deprecation warning in 3.1.9 and now we remove it completely. That way, all options for ray are passed in `ray_options` and for optuna in `optuna_options`. Since `cpus` and `gpus` were only for Ray, it makes sense to merge them in `ray_options` and remove them from the init params. I can raise a TypeError to make it more explicit and add back the params to avoid breaking the API.

elephaint: Ok, personally I think it's maybe a bit fast to deprecate & remove in such a short period (e.g. remove them after a year or so? that gives users the time to migrate), so agree on adding it back for now and raising TypeError

elephaint (approved):
Nice, thanks! I think the tests are breaking, happy to approve again once they succeed

PR #1545, opened 2026-06-02
[FIX] Allow CV without passing a df

### LINKED ISSUE #800: [Core] CV fails when df is not provided

marcopeix:
Currently, cross-validation cannot run without passing a dataframe. Now, we can reuse the saved data from fitting the model.

### DIFF
--- a/neuralforecast/core.py
+++ b/neuralforecast/core.py
@@ -1941,8 +1941,10 @@ class NeuralForecast:
             # Add original input df's y to forecasts DataFrame
             if df is None:
                 # Reconstruct the target from the stored dataset. The dataset's
-                # temporal values are scaled, so undo any target scaling
-                target_values = self.dataset.temporal[:, self.dataset.y_idx].numpy()
+                # temporal values are scaled, so undo any target scaling.
+                target_values = (
+                    self.dataset.temporal[:, self.dataset.y_idx].clone().numpy()
+                )
                 if self.scalers_:
                     target_values = self._scalers_target_inverse_transform(
                         target_values.reshape(-1, 1), self.dataset.indptr
--- a/neuralforecast/core.py
+++ b/neuralforecast/core.py
@@ -1939,6 +1939,21 @@ class NeuralForecast:
             fcsts_df = ufp.horizontal_concat([fcsts_df, fcsts])
 
             # Add original input df's y to forecasts DataFrame
+            if df is None:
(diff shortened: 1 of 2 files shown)


neuralforecast/core.py
```diff
@@ -1939,6 +1939,21 @@ def _no_refit_cross_validation(
             fcsts_df = ufp.horizontal_concat([fcsts_df, fcsts])
 
             # Add original input df's y to forecasts DataFrame
+            if df is None:
+                # Reconstruct the target from the stored dataset. The dataset's
+                # temporal values are scaled, so undo any target scaling
+                target_values = self.dataset.temporal[:, self.dataset.y_idx].numpy()
```
nasaul: [`_scalers_target_inverse_transform`](https://github.com/Nixtla/neuralforecast/blob/10e8f5ed3ec58c193cacdd0c439420d4aff69cdf/neuralforecast/core.py#L331) transforms data in-place. I think we should add `.clone().numpy()` just to be safe for that the data remains the same after calling cv without df.

nasaul (approved):
Great job Marco, just a nit before merging

neuralforecast/core.py
marcopeix: Good catch, indeed!

elephaint:
There's minor issues with this PR still:

1. Column names are wrong when df is None and CV is called without re-passing col kwargs.

If a user passes non-standard id / time / target col names, they get overridden in the output to generic names because they aren't passed down from self.id_col, self.time_col, self.target_col. 

2. The scaler inverse-transform path is untested; NHITS is built with no local_scaler_type in the test.

3. A minor nit is that Saul's addition unnecessarily creates a copy of the data on the non-scaler path (his remark is only applicable on the scaler path, so we don't need to create a copy if there is no scaler, which is the majority of cases)

PR #1548, opened 2026-06-04
[FIX] Further fixes for CV with no df

marcopeix:
- Prevents the override of non-default column names
- Avoid `clone()` when no local scaler is used
- More thorough test


elephaint (approved):
Nice, thanks!

PR #1552, opened 2026-06-10
[FEAT] Support categorical exog for univariate models

### LINKED ISSUE #916: [Core] Categorical Features

According to the [Exogenous Variables Tutorial](https://nixtlaverse.nixtla.io/neuralforecast/examples/exogenous_variables.html):
```
When including exogenous variables always use a scaler by setting the scaler_type hyperparameter. The scaler will scale all the temporal features: the target variable y, historic and future variables.
```
That means that all non-static variables are scaled and treated as real numbers. Neural-Forecast should give the option of specifying categorical features that should be embedded and non-categorical features that should be scaled. Using the example from the tutorial `gen_forecast` should be scaled, while `week_day` should be treated as a categorical.

### Versions / Dependencies

1.6.4

### Reproduction script

https://nixtlaverse.nixtla.io/neuralforecast/examples/exogenous_variables.html

### Issue Severity

Medium: It is a significant difficulty but I can work around it.

### LINKED ISSUE #1098: Categorical Features for TFT
### Description

I guess an actual TFT can handle categorical features with embeddings, does nixtla have something similar or it always needs to be done manually with label encodings?

### Use case

I am trying to put string in in the input as feature but the model is not being able to parse it.

marcopeix:
Phase 1 for supporting exogenous features in NeuralForecast.

**How it works**:
- Categorical features are treated through a learned embedding
- The dimension of the embedding can be set manually or we can use a heuristic like "fastai", "sqrt", or "half" to take into account the cardinality of the categorical features
- Introduce the `categorical_cardinalities` argument which is dict that lists the cardinality of each category. Similar function and usage as `n_series` for multivariate models. We could infer it from the data, but it's a much larger change.
- Adds the flag `CAT_EXOGENOUS` to `BaseModel`.
- The learned embedding also has a reserved axis for unknown categories at the time of predicting
- **Only univariate non-recurrent models support categorical features for now**

**What is excluded in this first phase**:
- Recurrent models
- Multivariate models
- Explainability is disabled if categorical features are detected
- Simulations are disabled if categorical features are detected

**Small experiment**
NHITS on AirPassengers, forecasting the last 12 months
Categorical feature: month

| unique_id | MAE (label-encoded) | MAE (embeddings) | improvement % |
|-----------|--------------------:|-----------------:|--------------:|
| Airline1  | 11.160              | 10.011           | 10.301        |
| Airline2  | 11.160              | 10.011           | 10.301        |
| overall   | 11.160              | 10.011           | 10.301        |

**Usage example**
```python
model = NHITS(
        h=h,
        input_size=INPUT_SIZE,
        max_steps=MAX_STEPS,
        scaler_type="robust",
        futr_exog_list=["month"],
        cat_exog_list=["month"],
        categorical_cardinalities={"month": 12},
)
nf = NeuralForecast(models=[model], freq="ME")
```


elephaint (changes requested):
Nice! This is a good start, and great that this is finally coming to NF 👍 

Initial review:
- I'd name it EXOGENOUS_CAT, to be consistent with previous naming
- Cardinality should follow from the data, and how the cardinality is baked into the model follows from the "cat_emb_dim". In the current code, I can supply "month":10 and what happens? Which is weird because it should be month:12, so if the data shows higher cardinality, that should supersede, or at least we should have a mechanism to deal with that. 
- What about missing values? I'd add +1 on the cardinality.
- What about new values (unseen during fit)? Again probably add +1 on the cardinality. See below also for training the OOV slot.
- What about embedding stat_exog? Often the largest source of categorical exogenous features. (Nearly every stat_exog == categorical exog, there is almost never an ordinal ranking in stat_exog)

Some bugs worth addressing:
- `val_df` is never encoded, so there's a mismatch
- cross_validation endpoint has some silent bugs; categorical_vocab_ needs to be reset. Full CV is encoded before test_size is held out.
- Conflicting cross-model cardinalities are now silently addressed, but they should be addressed by having the cardinality follow from the data.
- The unseen position is never trained so its slot stays at random init (not very opinionated on this, but why random? E.g. why not an "average" embedding?)

marcopeix:
@elephaint 
> Cardinality should follow from the data, and how the cardinality is baked into the model follows from the "cat_emb_dim". In the current code, I can supply "month":10 and what happens? Which is weird because it should be month:12, so if the data shows higher cardinality, that should supersede, or at least we should have a mechanism to deal with that.

To this point, I designed it like that by choice. It mirrors the current behavior of `n_series`. For `n_series`, users must pass the correct number, and we actually never verify against the data, and it's not inferred from the data.

I agree it's not ideal. And to your point, we do validate if the cardinality is lower than the actual cardinality. On the other hand, if a user over-declares (say 20 months), then we just waste embedding dimensions. So the mechanism if already in place to supersede the user input if the supplied cardinality is lower than the observed one.

So for now, I suggest we keep it that way. The ideal scenario is to have NF build it from the supplied training data, but that's a bigger refactor, I don't want to blow up this already big PR.

marcopeix:
@elephaint 
> What about missing values? I'd add +1 on the cardinality.

This is deferred because we currently reject exogenous columns with NaN. Need to think about how we validate exog in general before doing this. Again, I don't think we should do it in this PR.

marcopeix:
> What about new values (unseen during fit)? Again probably add +1 on the cardinality. See below also for training the OOV slot.

Already in place. The first index is reserved for unseen values during fit.

neuralforecast/core.py
elephaint, line 1970: Add categorical_vocab_ to the snapshot.

neuralforecast/core.py
```diff
@@ -541,6 +630,15 @@ def fit(
         self._cs_df: Optional[DataFrame] = None
         self.prediction_intervals: Optional[PredictionIntervals] = None
 
+        # Categorical vocabulary is rebuilt for each fit; built once from the
+        # training panel in `_prepare_fit` and reused at predict time.
+        self.categorical_vocab_ = {}
```
elephaint: I think df=None breaks this

neuralforecast/common/_base_model.py
```diff
@@ -474,14 +544,82 @@ def _check_exog(self, dataset):
...
+        cont_idxs = [i for i, c in enumerate(exog_list) if c not in cat_exog_list]
+        pieces = []
+        if cont_idxs:
+            pieces.append(x[..., cont_idxs])
+        for emb, col in zip(embeddings, cat_exog_list):
+            pos = exog_list.index(col)
+            pieces.append(self._lookup_embedding(emb, x[..., pos].long()))
+        return torch.cat(pieces, dim=-1)
+
+    def _lookup_embedding(self, emb, idx):
```
elephaint, line 605: Gate it on self.training / compute once at inference, otherwise unnecessarily costly. And only need to average observed rows.

elephaint (changes requested):
See comments in code. Other remarks:

- Not sure auto models are correctly handled; cat_exog_list in an Auto config isn't detected I think.
- prediction_intervals truncates the vocab: the internal CV rebuilds the vocab from df-minus-test_size and the final models train with it; categories only in the tail become permanent OOV, unlike plain fit.
- Test gap: mixed continuous + categorical in one stream currently untested

elephaint (approved):
👍

PR #1553, opened 2026-06-11
Translate `input_size_multiplier` in user-supplied Auto configs

### LINKED ISSUE #571: Default configs of AutoNHITS seems invalid
AutoNHITS.default_config is not a valid config if provided as argument since it uses 'input_size_multiplier', which is only translated to the necessary 'input_size' if config==None.
This makes configs that are only slight modifications of the default unnecessarily cumbersome

After a rough scan of the code this seems to apply for most or all AutoModels

Version 1.5.0

Vedant-Agarwal:
Fixes #571

## What

The `Auto*` models' `default_config` expresses input sizes as horizon multiples (`input_size_multiplier`, `inference_input_size_multiplier`), but `BaseAuto` only translated those keys into concrete `input_size` / `inference_input_size` values on the `config=None` path. A config built on top of `default_config` — copy it, tweak a value, pass it back — failed at construction with:

```
ValueError: `config` is missing required parameter(s) for MLP: ['input_size']
```

This applies the same translation `get_default_config` performs to user-supplied configs as well, for both backends:

- ray dict configs: translated once in `BaseAuto.__init__` (list multipliers become `tune.choice([h * x, ...])`, mirroring the default-config behavior)
- optuna config functions: translated per-trial inside the wrapper (scalar multiplier becomes `h * multiplier`)

Explicit `*input_size` keys always take precedence; multiplier keys are removed so they never reach the model constructor. No behavior change for configs that don't use multiplier keys.

## Tests

Two new tests in `tests/test_common/test_base_auto.py`, both referencing #571:

- `test_default_config_copy_is_valid_user_config` — copies `AutoMLP.default_config`, modifies values, fits and predicts via ray
- `test_optuna_config_translates_input_size_multiplier` — optuna config function suggesting a scalar `input_size_multiplier`, fits and predicts

Without the source change both fail with the exact error from #571; with it:

```
$ pytest tests/test_common/test_base_auto.py
18 passed
```

`pre-commit run` (ruff, mypy) passes on the changed files.

### DIFF
--- a/neuralforecast/common/_base_auto.py
+++ b/neuralforecast/common/_base_auto.py
@@ -198,6 +198,10 @@ class BaseAuto(pl.LightningModule):
             raise ValueError(
                 f"Unknown backend {backend}. The supported backends are 'ray' and 'optuna'."
             )
+        # Translate `*input_size_multiplier` entries (as used by the models'
+        # `default_config`) into concrete `*input_size` ones, so that configs
+        # built on top of `default_config` are valid `config` arguments.
+        config_base = self._translate_input_size_multipliers(config_base, h)
 
         # Shallow-copy user-supplied options so subsequent mutations
         # (default resolution) don't leak back to the caller.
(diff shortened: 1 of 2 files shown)


marcopeix (approved):
Thank you for the great work!

Issue #1100, opened 2024-08-09
All Multivariate Models (Except SOFTS) Cannot Be Ran


Teculos:
### What happened + What you expected to happen

I'm seeing multiple issues (all related to matrix dimensions it seems) for all multivariate models (Except HINT because I could not determine S parameter from documentation and SOFT which seems to work). This is reproducible in both standard models and Auto models.

Errors presented are not the full stack but reduced for cleanliness:

**TSMixer**
```
  File "/burg/pmg/users/aec2244/mambaforge/torch/lib/python3.10/site-packages/neuralforecast/models/tsmixer.py", line 118, in forward
    x = x * self.weight
RuntimeError: The size of tensor a (4) must match the size of tensor b (7) at non-singleton dimension 2
```
**TSMixerx**

```
  File "/burg/pmg/users/aec2244/mambaforge/torch/lib/python3.10/site-packages/neuralforecast/models/tsmixerx.py", line 146, in forward
    x = x * self.weight
RuntimeError: The size of tensor a (4) must match the size of tensor b (7) at non-singleton dimension 3
```

**TimeMixer**
```
  File "/burg/pmg/users/aec2244/mambaforge/torch/lib/python3.10/site-packages/neuralforecast/models/timemixer.py", line 85, in _normalize
    x = x * self.affine_weight
RuntimeError: The size of tensor a (4) must match the size of tensor b (7) at non-singleton dimension 2
```

**StemGNN**

```
  File "/burg/pmg/users/aec2244/mambaforge/torch/lib/python3.10/site-packages/neuralforecast/models/stemgnn.py", line 339, in self_graph_attention
    key = torch.matmul(input, self.weight_key)
RuntimeError: mat1 and mat2 shapes cannot be multiplied (175x4 and 7x1)
```

**MLPMultivariate**

```
  File "/burg/pmg/users/aec2244/mambaforge/torch/lib/python3.10/site-packages/torch/nn/modules/linear.py", line 117, in forward
    return F.linear(input, self.weight, self.bias)
RuntimeError: mat1 and mat2 shapes cannot be multiplied (25x384 and 672x1024)
```

### Versions / Dependencies

neuralforecast is 1.7.4 
datasetsforecast is 0.0.8
pytorch_lightning is 2.3.0
torch is 2.4.0+cu121

### Reproduction script

Reproduced the error with only Nixtla related packages, including NHITS

```

import optuna
import pandas as pd

from neuralforecast import NeuralForecast
[...]

Issue #1291, opened 2025-03-18
Auto models get ray.exceptions.ActorDiedError when run on multi-GPU node


bstewart311:
### What happened + What you expected to happen

I started a multi-gpu node within the AWS SageMaker JupyterLab environment, then git cloned neuralforecast, navigated to the experiments/long_horizon directory, created a conda long_horizon environment using the environment.yml command (as described in the long_horizon/readme.md).  Then I ran:

- python run_nhits.py --dataset 'ETTh1' --horizon 96 --num_samples 1

I got the following error:
--------------------------------------
(_train_tune pid=22946) [rank: 1] Child process with PID 23026 terminated with code 1. Forcefully terminating all other processes to avoid zombies 🧟
(raylet) A worker died or was killed while executing a task by an unexpected system error. To troubleshoot the problem, check the logs for the dead worker. RayTask ID: ffffffffffffffff14b6988a6a4935518fd4985a01000000 Worker ID: fa49156216bca352caee735419dc8cf5a59f668a96d20f17a94ddec9 Node ID: 6eabca3b606628a676f68efc346d601b765017331b5032e290b6b99e Worker IP address: 169.255.255.2 Worker port: 33095 Worker PID: 22946 Worker exit type: SYSTEM_ERROR Worker exit detail: Worker unexpectedly exits with a connection error code 2. End of file. There are some potential root causes. (1) The process is killed by SIGKILL by OOM killer due to high memory usage. (2) ray stop --force is called. (3) The worker is crashed unexpectedly due to SIGSEGV or other unexpected errors.
2025-03-18 19:36:13,639 ERROR tune_controller.py:1331 -- Trial task failed for trial _train_tune_39116_00000
Traceback (most recent call last):
  File "/home/sagemaker-user/.conda/envs/long_horizon/lib/python3.11/site-packages/ray/air/execution/_internal/event_manager.py", line 110, in resolve_future
    result = ray.get(future)
             ^^^^^^^^^^^^^^^
  File "/home/sagemaker-user/.conda/envs/long_horizon/lib/python3.11/site-packages/ray/_private/auto_init_hook.py", line 21, in auto_init_wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/sagemaker-user/.conda/envs/long_horizon/lib/python3.11/site-packages/ray/_private/client_mode_hook.py", line 103, in wrapper
    return func(*args, **kwargs)
[...]

Issue #1037, opened 2024-06-13
[common] Issue with multi-gpu and `ddp_spawn` strategy when running predict


matthieuhumeau:
### What happened + What you expected to happen

The `predict` method fails with the following error when the model has been trained on multi-gpu with `ddp_spawn` strategy:
`TypeError: vstack(): argument 'tensors' (position 1) must be tuple of Tensors, not NoneType`

This seems to be an issue with the PyTorch Lightning Trainer returning `None` when calling [`predict`](https://lightning.ai/docs/pytorch/stable/common/trainer.html#predict) with multi-gpu. Looks like there is already an existing fix for this: https://github.com/Nixtla/neuralforecast/pull/391/files
But the issue persists on my side. I was able to resolve it by modifying `common/_base_windows.py` to drop the `strategy` argument from my `trainer_kwargs`. 

I'm using:
```
trainer_kwargs = {
        'accelerator': 'gpu',
        'devices': 8, 
        'strategy': 'ddp_spawn'  # Distributed Data Parallel strategy
    }
```

### Versions / Dependencies

Running this on Sagemaker (AL2, `5.10.215-203.850.amzn2.x86_64`)
Python 3.10
`torch==2.1.0`
`pytorch-lightning==2.2.5`
`neuralforecast==1.7.2`

### Reproduction script

from utilsforecast.data import generate_series
from neuralforecast import NeuralForecast
from neuralforecast.models import NBEATS 
from neuralforecast.losses.pytorch import DistributionLoss
import torch
torch.set_float32_matmul_precision('high')

def main():
    series = generate_series(10, min_length=200, max_length=500)
    h = 7
    valid = series.groupby('unique_id', observed=True).tail(h)
    train = series.drop(valid.index)
    
    trainer_kwargs = {
            'accelerator': 'gpu',
            'devices': 8,
            'strategy': 'ddp_spawn'}

    models = NBEATS(h=h,
                    input_size=7,
                    loss=DistributionLoss(distribution='Poisson', level=[90]),
                    max_steps=100,
                    scaler_type='standard',
                    **trainer_kwargs)

    model = NeuralForecast(models=[models], freq='D', )
    model.fit(train)
    
    p = model.predict(train)

if __name__ == "__main__":
    main()

### Issue Severity

Medium: It is a significant difficulty but I can work around it.

jmoralez:
[...]

Issue #686, opened 2023-07-07
Code stuck on "initalizing ddp" when using more than one gpu on neuralforecast AutoTFT, AutoNHITs


philip-ndikum:
### What happened + What you expected to happen

When running this notebook with multi-GPU 

https://colab.research.google.com/github/Nixtla/neuralforecast/blob/main/nbs/examples/IntermittentData.ipynb

Code stuck on "initalizing ddp" - are there any parameters to control the number of GPU's utilized? Example if I have 10 GPUs but I only want to use one GPU for AutoTFT how could I do this without manually changing the low-level code? Could I define this globally in code or notebook?

Related to this issue here: https://github.com/Lightning-AI/lightning/issues/4612

Error occurs here: https://github.com/Nixtla/neuralforecast/blob/main/neuralforecast/losses/pytorch.py

### Versions / Dependencies

neuralforecast==1.5.0

### Reproduction script

nf.fit(df=Y_df)

### Issue Severity

High: It blocks me from completing my task.

TDL77:
I have the same problem when training just with neuralforecast , without using ray.
There are 2 video cards and I prescribe in the network parameters:
             strategy='ddp_notebook', # 'dp'|'ddp_notebook'/'ddp_spawn'
             accelerator = 'gpu',
             devices = [0,1],
, and constantly error You selected `Trainer(strategy='')` but process forking is not supported on this platform.

ValueError: You selected `Trainer(strategy='ddp_notebook')` but process forking is not supported on this platform. We recommed `Trainer(strategy='ddp_spawn')` instead.

ValueError: You selected `Trainer(strategy='ddp_fork')` but process forking is not supported on this platform. We recommed `Trainer(strategy='ddp_spawn')` instead.

MisconfigurationException: `Trainer(strategy='ddp_spawn')` is not compatible with an interactive environment. Run your code as a script, or choose one of the compatible strategies: `Fabric(strategy='dp'|'ddp_notebook')`. In case you are spawning processes yourself, make sure to include the Trainer creation inside the worker function.

marcopeix:
This is not an issue anymore. Not sure when it was solved, but the latest version (v3.2.0) doesn't have this issue.

You can do something like:
```python
from neuralforecast.auto import AutoTFT

config = AutoTFT.get_default_config(h=12, backend="ray")
config["accelerator"] = "gpu"
config["devices"] = [0]
model = AutoTFT(h=12, config=config)
```

[...]

PR #1555, opened 2026-06-12
Fail fast when multivariate models are configured with multiple devices

Vedant-Agarwal:
Addresses #1100 (and the duplicate reports in #1291, #1037 and #686).

## What

Multivariate models currently crash with cryptic tensor-shape `RuntimeError`s deep inside training when configured with more than one device — because data-parallel strategies split each batch across devices, while multivariate models require all series to be present in every batch (as @jmoralez diagnosed in #1100).

This PR makes that failure mode explicit: model construction (and the Spark `_fit_distributed` path) now raises a clear `ValueError` when a multivariate model is configured with multiple devices, telling the user exactly why and how to fix it (`devices=1`, or `trainer_kwargs={'devices': 1}` for Auto models). `predict` already self-protected by forcing a single device; `fit` was the unguarded path.

Full multi-GPU support for multivariate models would be the real resolution of #1100 and remains future work — this PR turns an opaque crash into a documented limitation in the meantime.

## Note on SOFTS

The issue title says "except SOFTS", but the thread only establishes that SOFTS *doesn't crash* — under batch-splitting it would still compute cross-series interactions over an incomplete subset of series, which is silent incorrectness rather than support. Since SOFTS declares `MULTIVARIATE = True`, it is guarded too. Happy to exempt it if maintainers prefer.

## Details

- `_num_requested_devices()` handles the device-spec forms (`int`, `-1`, list, `"auto"`, numeric strings) and multiplies by `num_nodes`
- The check runs at `BaseModel.__init__` right after accelerator/devices resolution — before any `Trainer` exists, so it fires immediately and needs no GPU
- New CPU-only tests (33) parametrized over TSMixer / MLPMultivariate / SOFTS × device specs, with univariate models as negative controls; tests fail without the source change
- `tests/test_common/` plus the multivariate model test files: 127 passed, no regressions; pre-commit (ruff, mypy) clean

### DIFF
--- a/neuralforecast/common/_base_model.py
+++ b/neuralforecast/common/_base_model.py
@@ -87,6 +87,57 @@ def tensor_to_numpy(tensor: torch.Tensor) -> np.ndarray:
     return tensor.numpy()
 
 
+def _num_requested_devices(trainer_kwargs: dict) -> int:
+    """Best-effort count of the devices a ``pl.Trainer`` built from
+    ``trainer_kwargs`` would use. Returns 1 when the count cannot be
+    determined."""
+    devices = trainer_kwargs.get("devices")
+    if devices is None:
+        return 1
+    if isinstance(devices, (list, tuple)):
+        return len(devices)
+    if isinstance(devices, str):
+        if devices.strip() == "auto":
+            devices = -1
+        else:
+            try:
+                devices = int(devices)
+            except ValueError:
+                return 1
+    if devices == -1:
+        # all devices of the accelerator. only CUDA can resolve to more than
+        # one device here, MPS is single-device and Lightning's CPU
+        # accelerator defaults to a single process.
+        accelerator = trainer_kwargs.get("accelerator")
+        if accelerator in (None, "auto", "gpu", "cuda"):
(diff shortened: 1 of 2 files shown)


marcopeix:
It prevents the issue but doesn't fix it. Let's see if we can find an actual fix.

Vedant-Agarwal:
@marcopeix - Multivariate models break on multiple devices because they need every batch to contain all the series, and data-parallel training splits each batch across devices. Real multi-device support would need custom distributed batching or model parallelism — a large change to the base model.

A smaller option that just works: instead of raising, fall back to a single device (with a warning) for multivariate models, so the run succeeds instead of crashing. I prototyped this locally and unit-tested it — it coerces devices=2 / [0,1] / "2" and multi-node configs down to a single device and warns, leaving single-device runs unchanged.

Would you prefer this auto-fallback, or do you want to pursue real multi-device support? Happy to push whichever you prefer.

marcopeix:
#1557 fixes #1100, this PR is not needed. Also, issues listed above are different from #1100, they are not duplicates.

PR #1560, opened 2026-06-20
[FEAT] Add GELU as a grn_activation option in TFT

autokaiai:
## What

Adds `GELU` (`torch.nn.functional.gelu`) as a selectable value for the TFT `grn_activation` argument. It is a single new entry `"GELU": F.gelu` in the `activation_map` of `get_activation_fn`.

## Why

This is a completeness / consistency change, not a performance claim. #1175 made the TFT GRN activation configurable and exposed a fixed set of standard `torch.nn.functional` activations: `ReLU, Softplus, Tanh, SELU, LeakyReLU, Sigmoid, ELU, GLU`. `GELU` is a standard, widely used activation that was simply absent from that set. Adding it brings the option list to parity with the activations users commonly expect. The change is purely additive, so existing behaviour and defaults are unchanged.

## Changes

- `neuralforecast/models/tft.py`: add `"GELU": F.gelu` to `activation_map`; append `'GELU'` to the `grn_activation` docstring list.

## Verification

- Existing `tests/test_models/test_tft.py` passes (2 passed).
- Repo pre-commit hooks (`ruff`, `mypy`) pass on the change.
- End-to-end: trained a TFT with `grn_activation="GELU"` and produced valid forecasts.

Refs #1175


tests/test_models/test_tft.py
```diff
@@ -9,6 +12,12 @@ def test_tft_model(suppress_warnings):
     check_model(TFT, ["airpassengers"])
 
 
+def test_grn_activation_gelu_lookup():
+    # "GELU" must resolve to F.gelu, not silently fall back to F.elu
+    # (get_activation_fn uses activation_map.get(name, F.elu)).
+    assert get_activation_fn("GELU") is F.gelu
+
+
```
marcopeix: This change if very straightfoward. We don't need to add a test for it. At least, not this test in particular. It's just asserting it's in the list. We could a test that trains a TFT model for 2 steps for each grn_activation option, but I think it's overkill.

marcopeix (changes requested):
@autokaiai Thanks for the contribution. See my comment.

tests/test_models/test_tft.py
autokaiai: Thanks for the quick review, @marcopeix! Makes sense, so I've removed the test. The change is now just the `activation_map` entry and the docstring update. This is my first open source contribution, so I appreciate the guidance. 🙏

marcopeix (approved):
Thanks for the contribution!

PR #1570, opened 2026-07-14
[FIX] Fix AutoHINT ray serialization issue

### LINKED ISSUE #559: Ray error with AutoHINT
I'm trying to use AutoHINT in this way:

```
model = AutoHINT(
    NHITS,
    h=horizon,
    loss=GMM(n_components=10, quantiles=quantiles),
    valid_loss=sCRPS(quantiles=quantiles),
    S=S_df.values,
    config=config,
    search_alg=HyperOptSearch(),
    num_samples=2,
    verbose=True,
)
```

But I get this error:

`ValueError: The actor ImplicitFunc is too large (235 MiB > FUNCTION_SIZE_ERROR_THRESHOLD=95 MiB). Check that its definition is not implicitly capturing a large array or other object in scope. Tip: use ray.put() to put large objects in the Ray object store.`

The same `config` (without the "reconciliation" parameter) works if I use AutoNHITS instead of AutoHINT.

marcopeix:
Previously, Ray would try to serialize the entire summation matrix, causing serialization errors. We now use an `ObjectRef` to prevent serializing a very large object and confirm AutoHINT runs end-to-end.

### DIFF
--- a/tests/test_models/test_hint.py
+++ b/tests/test_models/test_hint.py
@@ -167,14 +167,11 @@ def test_auto_hint_large_S_not_captured_in_trainable():
     captured_size = len(cpickle.dumps(auto._train_tune))
     assert captured_size > RAY_FUNCTION_SIZE_THRESHOLD
 
-    # Once `S` lives in the object store, `self` no longer
-    # holds the ndarray directly. Ray captures the ObjectRef by reference in the
-    # actual actor-registration path, so the array is not re-serialized per trial.
     if not ray.is_initialized():
         ray.init()
     auto.S = ray.put(S)
-    assert isinstance(auto.S, ray.ObjectRef)
-    assert not isinstance(auto.S, np.ndarray)
+    recaptured_size = len(cpickle.dumps(auto._train_tune))
+    assert recaptured_size < RAY_FUNCTION_SIZE_THRESHOLD
 
 
 # `AutoHINT.fit` must move `S` into the Ray object store before the trainable is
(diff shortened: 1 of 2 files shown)


tests/test_models/test_hint.py
```diff
@@ -121,3 +130,102 @@ def test_hint_fit_predict():
...
+    captured_size = len(cpickle.dumps(auto._train_tune))
+    assert captured_size > RAY_FUNCTION_SIZE_THRESHOLD
+
+    # Once `S` lives in the object store, `self` no longer
+    # holds the ndarray directly. Ray captures the ObjectRef by reference in the
+    # actual actor-registration path, so the array is not re-serialized per trial.
+    if not ray.is_initialized():
+        ray.init()
+    auto.S = ray.put(S)
+    assert isinstance(auto.S, ray.ObjectRef)
```
nasaul: nit: this assertion is trivially true given the `ray.put` on the line above, so it tests Ray rather than the fix. Consider also asserting that the trainable pickle is now small, e.g. `assert len(cpickle.dumps(auto._train_tune)) < RAY_FUNCTION_SIZE_THRESHOLD`, so the test actually guards against the array being re-captured.

nasaul (approved):
LGTM! The fix is correct and well-scoped, left one small nit on the test file.

tests/test_models/test_hint.py
marcopeix: Will adjust, thanks!

PR #1571, opened 2026-07-14
[FIX] Fix loss computation in multi-device training

### LINKED ISSUE #1558: Validation loss computation in multi-device training
Currently, validation loss computation is not adapted for multi-device training.

For univariate models, the reported validation loss is only partial (we only use the loss from the first device).

For multivariate models, the same loss is computed many times, which is inefficient.

Better to also distribute the validation step, and correctly combine and weigh the loss.

marcopeix:
Ensures that in multi-device training, the reported loss is not computed many times for multivariate models, and we correctly weigh the loss for univariate models.

### DIFF
--- a/neuralforecast/common/_base_model.py
+++ b/neuralforecast/common/_base_model.py
@@ -840,11 +840,12 @@ class BaseModel(pl.LightningModule):
         # batches; all_gather then combines the ranks.
         stacked = torch.stack(self.validation_step_outputs)
         totals = self.all_gather(stacked.sum(dim=0)).reshape(-1, 2).sum(dim=0)
-        avg_loss = (totals[0] / totals[1]).item()
+        loss_sum, count = totals.tolist()
+        avg_loss = loss_sum / count
         self.log(
             "ptl/val_loss",
             avg_loss,
-            batch_size=int(totals[1].item()),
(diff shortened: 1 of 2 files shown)


neuralforecast/common/_base_model.py
```diff
@@ -835,13 +835,16 @@ def set_test_size(self, test_size):
...
+        # batches; all_gather then combines the ranks.
+        stacked = torch.stack(self.validation_step_outputs)
+        totals = self.all_gather(stacked.sum(dim=0)).reshape(-1, 2).sum(dim=0)
+        avg_loss = (totals[0] / totals[1]).item()
         self.log(
             "ptl/val_loss",
             avg_loss,
-            batch_size=losses.size(0),
-            sync_dist=True,
+            batch_size=int(totals[1].item()),
```
elephaint: Is there are better way of getting the batch_size here? Because doing .item() triggers CPU-GPU synchronization

elephaint (approved):
Nice, non-blocking performance comment

neuralforecast/common/_base_model.py
marcopeix: Sure, used another way, slightly faster :)

PR #1573, opened 2026-07-16
[FEAT] Support cat feat for multivariate models

### LINKED ISSUE #1569: [FEAT] Categorical feature support for multivariate models
### Description

_No response_

### Use case

_No response_

marcopeix:
Extend support of categorical features to multivariate models

### DIFF
--- a/nbs/docs/capabilities/categorical_features.ipynb
+++ b/nbs/docs/capabilities/categorical_features.ipynb
@@ -36,9 +36,13 @@
    "metadata": {},
    "source": [
     "NeuralForecast handles categorical features with learned embeddings. Internally, it:\n",
+    "\n",
     "- builds a panel-wide vocabulary at `fit` time. This creates a mapping from category to index for each feature across all series. Index 0 is reserved for unseen categories during training.\n",
+    "\n",
     "- routes the categorical features around the scalers. Categorical features are not scaled, unlike numerical features.\n",
+    "\n",
     "- learns an embedding for each feature. One vector per category is created and is learned jointly with other continuous feaetures.\n",
+    "\n",
     "- concatenates each feature's embeddings with the continous features and feeds the result to the model.\n",
     "\n",
     "By using learned embeddings, models learn a rich representation of the features which usually results in better forecasts than using a traditional label encoder.\n",
@@ -51,7 +55,7 @@
    "id": "5bd16669",
    "metadata": {},
    "source": [
-    "### Choosing the embdding dimension"
+    "### Choosing the embedding dimension"
    ]
   },
   {
@@ -99,6 +103,9 @@
    "source": [
     "This is a large change that is being implemented in multiple phases. Currently, categorical features are support for:\n",
     "- univariate models\n",
+    "\n",
+    "- multivariate models\n",
+    "\n",
     "- non-recurrent models"
    ]
   },
@@ -116,9 +123,11 @@
    "metadata": {},
    "source": [
     "All elements listed here do not support the use of categorical features:\n",
+    "\n",
     "- recurrent models\n",
-    "- multivariate models\n",
+    "\n",
     "- explainability with categorical features\n",
+    "\n",
     "- simulation with categorical features"
    ]
   },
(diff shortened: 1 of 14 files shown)


tests/test_categorical.py
```diff
@@ -365,17 +372,76 @@ def test_recurrent_model_rejects_categoricals():
...
-        MLPMultivariate(
-            h=6,
-            input_size=12,
-            n_series=4,
-            max_steps=2,
-            hist_exog_list=["city"],
-            cat_exog_list=["city"],
-            categorical_cardinalities={"city": 4},
-        )
+@pytest.mark.parametrize("cls", [MLPMultivariate, TSMixerx])
```
nasaul: `XLinear` and `TimeXer` also get `EXOGENOUS_CAT = True` in this PR but aren't exercised by any test. `XLinear` supports all three streams, so it can simply be added to this parametrize list. `TimeXer` needs its own case (it has `EXOGENOUS_FUTR = False`), and it's the one most worth smoke-testing: it reshapes `hist_exog` into per-variate tokens (`[B, L, X*N]`), so each embedding dimension becomes a separate variate — it should work mechanically, but it's the least trivial consumer of the expanded feature axis.

neuralforecast/common/_base_model.py
```diff
@@ -1177,6 +1177,15 @@ def _parse_windows(self, batch, windows):
                     self.hist_cat_embeddings,
                 )
             else:
+                # [Ws, L, C, N] -> embed on the feature axis -> [Ws, C', L, N].
```
nasaul: Nit: this comment describes the result *after* the `swapaxes(1, 2)` on the line below — the embed + transpose expression itself produces `[Ws, L, C', N]`, and only then does `swapaxes` give `[Ws, C', L, N]`. Worth tightening so the next reader doesn't miscount a transpose. (Same in the futr branch below.)

Optional: this transpose–embed–transpose block is duplicated in the hist and futr branches; a small helper (or a `dim` argument on `_embed_stream`) would keep them in sync, though at two call sites it's a judgment call.

nbs/docs/capabilities/categorical_features.ipynb
```diff
@@ -99,6 +103,9 @@
    "source": [
     "This is a large change that is being implemented in multiple phases. Currently, categorical features are support for:\n",
     "- univariate models\n",
+    "\n",
+    "- multivariate models\n",
```
nasaul, line 107: The supported list now reads "univariate models / multivariate models / non-recurrent models", which is redundant — the first two are subsumed by "all non-recurrent models". Also, since this file was already being touched for the "embdding" typo, the nearby pre-existing typos ("feaetures", "continous", "are support for") could be fixed in the same pass.

nasaul (approved):
Approving with a few minor suggestions inline — the main one is test coverage for `TimeXer` and `XLinear`, which get `EXOGENOUS_CAT = True` but aren't exercised by any test.

nasaul (approved):
LGTM! One minor note (non-blocking)

neuralforecast/common/_base_model.py
```diff
@@ -1177,6 +1177,15 @@ def _parse_windows(self, batch, windows):
                     self.hist_cat_embeddings,
                 )
             else:
+                # _embed_stream works on the last axis, so transpose the feature
+                # axis to last, embed, then transpose back: [Ws, L, C, N] ->
+                # [Ws, L, C', N]. The swapaxes below then gives [Ws, C', L, N].
+                hist_exog = self._embed_stream(
```
nasaul, line 1183: Note on the recurrent path that feeds into this same embedding logic (the univariate branch above): for recurrent models, `hist_exog[:, self.input_size:] = 0.0` zeroes the horizon *before* embedding. A `0.0` categorical index is the OOV slot, and `_lookup_embedding` maps OOV to the mean of the learned embeddings — so unknown-future hist categoricals silently get an "average category" fill rather than the "no signal" zeros continuous features get. Arguably a reasonable neutral fill, but it's an implicit semantic decision; a short comment at the zeroing site (or in `_lookup_embedding`) would make it read as intentional.

PR #1574, opened 2026-07-17
[FEAT] Support categorial features for simulation paths

marcopeix:
Adds support for simulation paths when using categorical features


neuralforecast/core.py
```diff
@@ -1644,6 +1647,9 @@ def simulate(
             )
             futr_df = fcsts_df
 
+        # Encode categoricals with the vocabulary fitted on the training data
+        # (the df-provided path already encodes via _prepare_fit).
+        futr_df = self._encode_categoricals(futr_df)
```
elephaint, line 1671: I think we should also add futr_df validation here similar to how it's done in predict

tests/test_categorical.py
elephaint, line 1: The tests only assert shapes and non-finiteness, but e.g. we don't test .simulate with a provided df (all tests are with df=None).

elephaint (changes requested):
Nice, two minor comments!

elephaint (approved):
🥇

PR #1579, opened 2026-07-21
Initialize static scalers in distributed prediction workers

### LINKED ISSUE #1578: AttributeError: 'NeuralForecast' object has no attribute 'static_scalers_'
https://github.com/Nixtla/neuralforecast/blob/0dc5c89baa530f4c067e9ccf8e22196db3de5026/neuralforecast/core.py#L958-L963

Distributed predict fails on the workers:

`AttributeError: 'NeuralForecast' object has no attribute 'static_scalers_'`

In `neuralforecast/core.py`, `_predict_distributed` the worker creates a fresh NeuralForecast and sets `nf.scalers_ = {}`, but not `nf.static_scalers_`, so `_scalers_transform` crashes on `if self.static_scalers_`:

```python
nf = NeuralForecast(models=models, freq=freq)
nf.id_col = id_col
nf.time_col = time_col
nf.target_col = target_col
nf.scalers_ = {}
nf.static_scalers_ = {}  # missing
nf._fitted = True
```

Expected: distributed predict works.

### Versions / Dependencies

3.2.0

### Reproduction script

```python
import datetime as dt
import tempfile

from pyspark.sql import SparkSession
from neuralforecast import DistributedConfig, NeuralForecast

spark = SparkSession.builder.master("local[1]").getOrCreate()

df = spark.createDataFrame(
    [("a", dt.datetime(2026, 1, 1) + dt.timedelta(d), 1.0) for d in range(30)],
    ["unique_id", "ds", "y"],
)

nf = NeuralForecast(models=[NHITS(h=7, input_size=7, max_steps=1, accelerator="cpu")], freq="D")

with tempfile.TemporaryDirectory() as tmp:
    nf.fit(df=df, distributed_config=DistributedConfig(partitions_path=f"{tmp}/p", num_nodes=1, devices=1))
    nf.predict(engine=spark).show()
```

### Issue Severity

Medium

ShreyanshGoyal:
Closes #1578.

## Summary

Initialize static_scalers_ on the temporary NeuralForecast instance created inside each distributed prediction worker.

## Root cause

The worker setup initialized scalers_ but omitted static_scalers_. The local prediction path saves and accesses both attributes before transforming the worker dataset, so distributed prediction failed with AttributeError before model inference began.

The regression test exercises the production worker callback through the Fugue transform boundary and verifies that both scaler dictionaries are initialized.

## Test plan

- [x] Reproduce the issue using the same state constructed by a distributed worker
- [x] Add a focused distributed-worker regression test
- [x] Confirm the regression fails without the source fix with the reported AttributeError
- [x] Run the focused test: 1 passed
- [x] Run Ruff checks and formatting checks on changed files


marcopeix (changes requested):
@ShreyanshGoyal 
The fix is good, thanks!

I notice we should also fix the `_simulate_distributed` function. The bug exists there too. So in `core.py` if you can also add the static_scalers line under `nf.scalers_ = {}` that would be great!

Also, please sign the CLA so we can merge this!

ShreyanshGoyal:
Added the matching static_scalers_ initialization to _simulate_distributed in commit 2050515. I also added a regression test that exercises the simulation worker callback and confirms both scaler dictionaries are initialized. The two focused distributed tests pass, and Ruff checks are clean.

marcopeix (approved):
Nice work, thanks!

PR #1580, opened 2026-07-21
[FIX] Upgrade pillow

marcopeix:
To fix this dependabot alert: https://github.com/Nixtla/neuralforecast/security/dependabot/80


JQGoh (approved):

uv.lock
```diff
@@ -2245,7 +2245,7 @@ wheels = [
 
 [[package]]
 name = "neuralforecast"
-version = "3.1.9"
+version = "3.2.0"
```
JQGoh, line 2248: @marcopeix I meant this update. Previously perhaps the uv sync did not work and get to update the neural forecast version

JQGoh (approved):

uv.lock
marcopeix: oh, weird...

PR #1602, opened 2026-09-03
fix: preserve infinite losses in weighted reductions

### LINKED ISSUE #1601: [Losses] _divide_no_nan turns an infinite loss into 0.0, so broken trials win Auto search

What happened:
A diverged trial reported loss = 0.0. Its predictions were ±inf, so the loss
should have been inf. Because 0.0 is the lowest possible value, the Auto
search treated the broken trial as the best one, kept it through every ASHA
rung, and returned it as the winner. No warning, no exception. best_config
stored the diverged configuration.

What I expected to happen:
A diverged trial should report inf, or at least a very large loss, so it
ranks worst in the search. The winner should be one of the trials that
actually converged.

### Versions / Dependencies

```
neuralforecast 3.2.1
torch 2.13.0+cu130
ray 2.57.0
python 3.12.3
Linux
```

### Reproduction script

```python
import torch

y = torch.tensor([[[10.0], [20.0]]])
print(MAE()(y=y, y_hat=torch.full_like(y, float("inf"))))

# actual:   tensor(0.)
# expected: tensor(inf)
```

### Issue Severity

kiwoongyoon:
Title: fix: preserve infinite losses in weighted reductions

## Summary

- Preserve signed infinities in `_divide_no_nan` instead of replacing them with zero.
- Ignore zero-weight entries before multiplying weighted losses, preventing masked infinities from producing `NaN`.
- Cover unmasked and masked positive/negative infinite predictions with an MAE regression test.

Fixes #1601

## Test evidence

- `pytest tests/test_losses/test_pytorch.py::test_MAE_infinite_predictions -q` — 1 passed
- `pytest tests/test_losses/test_pytorch.py -q` — 16 passed
- `ruff check neuralforecast/losses/pytorch.py tests/test_losses/test_pytorch.py` — passed
- `git diff --check` — passed

## Risks or notes for maintainers

`_weighted_mean` is shared by point losses, so zero-weight non-finite values are now ignored consistently wherever masks or zero horizon weights are used. No public API or dependency changes are included.

### DIFF
--- a/neuralforecast/losses/pytorch.py
+++ b/neuralforecast/losses/pytorch.py
@@ -33,13 +33,14 @@ def _divide_no_nan(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
     Auxiliary function to handle divide by 0
     """
     div = a / b
-    return torch.nan_to_num(div, nan=0.0, posinf=0.0, neginf=0.0)
+    return torch.nan_to_num(div, nan=0.0, posinf=float("inf"), neginf=float("-inf"))
 
 
 def _weighted_mean(losses, weights):
     """
     Compute weighted mean of losses per datapoint.
     """
+    losses = torch.where(weights != 0, losses, 0.0)
     return _divide_no_nan(torch.sum(losses * weights), torch.sum(weights))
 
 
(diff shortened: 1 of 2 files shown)


marcopeix:
@kiwoongyoon Thanks for raising the issue and drafting a PR. I think I have a more complete fix in #1603

marcopeix:
Closing since merging #1603

PR #1603, opened 2026-09-03
[FIX] Avoid silent divergence by reporting inf losses

### LINKED ISSUE #1601: [Losses] _divide_no_nan turns an infinite loss into 0.0, so broken trials win Auto search

marcopeix:
Fixes #1601.  

A diverged trial reported loss `0.0`, the lowest possible value, so it won every Auto search.

`_divide_no_nan` sanitized the quotient instead of guarding the denominator. Every `inf`/`nan` loss reduced to `0.0`.

## Changes

- `_divide_no_nan`: guard `b == 0` instead of scrubbing the result. Non-finite losses propagate; the deliberate `y == 0` skips in MAPE/SMAPE/MASE.
- `_weighted_mean`: drop zero-weight entries first, so a masked-out `inf` doesn't become `nan * 0`.
- `weighted_average`: `clamp(min=1.0)` under-normalized when total weight was below 1, and `if dim` treated `dim=0` as falsy.
- Validation reports a non-finite `ptl/val_loss` as `+inf` with a warning, so a diverged trial ranks worst instead of erroring the search.
Training keeps its NaN raise; `inf` only warns (mixed precision recovers).


nasaul (approved):
LGTM

PR #1607, opened 2026-09-07
[FEAT] Add future exogenous covariates support to iTransformer

FedeCarollo:
## Description
This PR completes the support for future exogenous covariates (`futr_exog`) in the `iTransformer` model.

### Context & Motivation
`iTransformer.__init__` already accepted `futr_exog_list` in its signature, indicating planned support for future covariates. However, the model did not yet implement the inverted tokenization or projection for future exogenous variables, and `EXOGENOUS_FUTR` was left disabled.

This PR finishes the integration cleanly within the inverted Transformer architecture:
1. **Horizon Slicing & Inverted Embedding:** Future covariates are extracted over the forecast horizon ($h$) and each feature per series is projected via `futr_embedding = nn.Linear(h, hidden_size)` into its own token of shape `[batch_size, n_series * futr_exog_size, hidden_size]`.
2. **Cross-Variate Attention:** Covariate tokens are concatenated with target series tokens `[enc_out, futr_tokens]` before passing through the inverted Transformer encoder (`TransEncoder`), allowing self-attention across variates and future drivers.
3. **Target Forecast Projection:** Only the first $N$ target series tokens (`enc_out[:, :N, :]`) are projected through `projector` to generate the forecast horizon, filtering out covariate tokens while preserving the standard output shape `[batch_size, h * outputsize_multiplier, n_series]`.
4. **Reversible Instance Normalization:** Reversible normalization (`use_norm`) is consistently applied to both target lookback sequences and future covariate slices, and inverted on output forecasts.

### Backward Compatibility
- 100% backward compatible: when `futr_exog_list=None`, behavior and outputs are identical to the original implementation.
- All existing tests (including `test_itransformer_model`) pass without changes.

### Tests
Added `test_itransformer_futr_exog` in `tests/test_models/test_itransformer.py` covering:
- Multivariate panel forecasting (2 series, 2 future exogenous features) with end-to-end `fit` and `predict(futr_df)`.
- Univariate panel forecasting (1 series, 1 future exogenous feature).
- Validation of output shapes and absence of `NaN` values.

### Linting
- `ruff`: no issues on modified files.
- `mypy`: 97 pre-existing errors across the codebase, none introduced by this PR (`neuralforecast/models/itransformer.py` is clean).

### DIFF
--- a/neuralforecast/models/itransformer.py
+++ b/neuralforecast/models/itransformer.py
@@ -29,7 +29,10 @@ class iTransformer(BaseModel):
         h (int): Forecast horizon.
         input_size (int): autorregresive inputs size, y=[1,2,3,4] input_size=2 -> y_[t-2:t]=[1,2].
         n_series (int): number of time-series.
-        futr_exog_list (str list): future exogenous columns.
+        futr_exog_list (str list, optional): future exogenous columns known ahead across the forecast horizon.
+            Each future covariate sequence over the forecast horizon `h` is embedded as an inverted token
+            via a linear projection layer, allowing cross-variable self-attention between target series and
+            future covariates.
         hist_exog_list (str list): historic exogenous columns.
         stat_exog_list (str list): static exogenous columns.
         exclude_insample_y (bool): the model skips the autoregressive features y[t-input_size:t] if True.
@@ -72,7 +75,7 @@ class iTransformer(BaseModel):
     """
 
     # Class attributes
-    EXOGENOUS_FUTR = False
+    EXOGENOUS_FUTR = True
     EXOGENOUS_HIST = False
     EXOGENOUS_STAT = False
     MULTIVARIATE = True
@@ -170,9 +173,15 @@ class iTransformer(BaseModel):
         self.use_norm = use_norm
 
         # Architecture
+        # Inverted embedding for target series: projects lookback length (input_size) -> hidden_size
         self.enc_embedding = DataEmbedding_inverted(
             input_size, self.hidden_size, self.dropout
         )
+
+        # Inverted embedding for future exogenous variables: projects horizon length (h) -> hidden_size
+        if self.futr_exog_size > 0:
+            self.futr_embedding = nn.Linear(h, self.hidden_size)
+            self.futr_dropout = nn.Dropout(self.dropout)
 
         self.encoder = TransEncoder(
             [
@@ -198,7 +207,17 @@ class iTransformer(BaseModel):
             self.hidden_size, h * self.loss.outputsize_multiplier, bias=True
         )
 
-    def forecast(self, x_enc):
+    def forecast(self, x_enc, futr_exog=None):
+        """Generate forecasts from lookback series and optional future covariates.
+
+        Args:
+            x_enc (torch.Tensor): Lookback target sequences of shape [batch_size, input_size, n_series].
+            futr_exog (torch.Tensor, optional): Future exogenous covariates of shape
+                [batch_size, futr_exog_size, input_size + h, n_series].
+
+        Returns:
+            dec_out (torch.Tensor): Forecast tensor of shape [batch_size, h * outputsize_multiplier, n_series].
+        """
         if self.use_norm:
             # Normalization from Non-stationary Transformer
             means = x_enc.mean(1, keepdim=True).detach()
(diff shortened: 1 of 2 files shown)


neuralforecast/models/itransformer.py
```diff
@@ -208,25 +227,41 @@ def forecast(self, x_enc):
...
+            # Reshape each feature for each series as a token: [B, N * F, h]
+            futr = futr.permute(0, 3, 1, 2).reshape(B_f, N_f * F_f, h_f)
+
+            if self.use_norm:
+                f_means = futr.mean(-1, keepdim=True).detach()
+                futr = futr - f_means
+                f_stdev = torch.sqrt(
+                    torch.var(futr, dim=-1, keepdim=True, unbiased=False) + 1e-5
+                )
+                futr /= f_stdev
```
marcopeix: I think we should remove it, otherwise it messes up the scaling of features.

neuralforecast/models/itransformer.py
```diff
@@ -208,25 +227,41 @@ def forecast(self, x_enc):
...
-        # N: number of variate (tokens), can also includes covariates
+        _, _, N = x_enc.shape  # B: batch_size, L: input_size, N: n_series
+
+        # Inverted embedding for target series: B L N -> B N E
+        enc_out = self.enc_embedding(x_enc, None)
+
+        # Inverted embedding for future exogenous variables
+        if self.futr_exog_size > 0 and futr_exog is not None:
+            if futr_exog.ndim == 3:
+                futr_exog = futr_exog.unsqueeze(-1)
```
marcopeix: This can be deleted, I don't think it's ever reached.

marcopeix (changes requested):
@FedeCarollo , thank you for the PR, see my comments below!

For the tests:
  - use_norm=False — currently untested with covariates, and it's a structurally different code path.
  - A probabilistic loss (MQLoss) with futr_exog_list, to pin the h * outputsize_multiplier reshape. I ran this manually and it's correct — [2, 4, 15] for h=4,
    n_series=3, MQLoss — but nothing in the suite locks it down.
  - An assertion that predictions actually change when futr_df changes. Right now the test would still pass if futr_tokens were silently dropped, which is the
    failure mode most worth guarding.

  Minor: import pandas as pd / from neuralforecast import NeuralForecast are function-local here while the rest of the file imports at module level; and the two
  halves are near-identical, so @pytest.mark.parametrize over (n_series, n_temporal_features) would cut it roughly in half.

Also, we drop the lookback period of future exogenous features. Is that intentional?

We also need to update the capabilities table to show that iTransformer support futr_exog.

FedeCarollo:
Hi @marcopeix, thank you for the review! I've addressed all the comments and pushed the changes. Here is a breakdown of the updates:

### Tests
- **Imports & Parametrization**: Moved all imports to the module level. I used `@pytest.mark.parametrize` over `(n_series, n_temporal_features)` with values `[(1, 1), (2, 3)]` (using `2, 3` to catch potential square matrix reshape bugs), cutting the code duplication.
- **Normalization & Loss Coverage**: Added `parametrize` for both `use_norm` (`[True, False]`) and `loss` (`[MAE(), MQLoss(level=[10, 90])]`). I decided to parameterize the loss directly in the main test to avoid code duplication and implicitly test all combinations (Univariate/Multivariate x Norm x Loss), which strictly locks down the `h * outputsize_multiplier` logic. However, if you prefer a dedicated separate test just for the probabilistic loss, let me know and I can easily split it!
- **Robustness Check on Future Tokens**: I've added a strict loop that iterates over each temporal column in `futr_exog_list`, alters its values, and asserts that the predictions change (`not preds_changed.equals(preds)`). 
  - *Side Note on `generate_series`: To implement this assert, I noticed a minor bug in `generate_series` within `utils.py`. Because `random.seed(seed)` is called inside the `for` loop, all temporal features get populated with the exact same identical values. Moreover, the values remain constant across time. Additionally, the function hardcodes these mock features as `CategoricalDtype` (unless I'm missing an alternative way to generate continuous mock data!). Since these constraints make it tricky to mutate the array directly, I used `.assign(**{col: test_df[col].astype(float) + 1.0})` to gracefully bypass the categorical typing and enforce the feature variations for the test.*

### Model Logic (`itransformer.py`)
- **Full Sequence Embedding**: You were absolutely right about the lookback period being dropped. I updated the `futr_embedding` linear layer to project the entire sequence (`input_size + h`), and removed the slicing logic in `forecast()`. The model now natively embeds both the historical and future portions of the exogenous covariates.
- **Code Cleanup**: As per your comments, I deleted the `futr_exog.ndim == 3` unsqueeze block and completely removed the dynamic `use_norm` scaling block applied to `futr` to prevent it from messing up the feature scales.

### Documentation
- Updated the capabilities table in `nbs/docs/capabilities/overview.ipynb` (switched `-` to `F` for `iTransformer`).

*P.S. While looking at the codebase, I noticed `EXOGENOUS_HIST = False`. Now that the inverted embedding logic for covariates is well established here, adding support for `hist_exog_list` would be very straightforward following the exact same logic. I am completely available to implement it in a follow-up PR if you think it's a valuable addition!*

marcopeix (approved):
Very nice, thanks for your work!

PR #1612, opened 2026-09-11
[FEAT] Add `decoder_activation`

### LINKED ISSUE #1608: Most models do not allow user to specify activation of the final layer?
### Description

E.g. for LSTM,

the final layer is fixed to activation="ReLU" in the code, but my application want "tanh", there is no easy way to change this.

And I check other models, and found most models do not allow user to specify activation of the final layer, can we add such option flag?

Thanks.

### Use case

_No response_

marcopeix:
Adds the `decoder_activation` kwarg for recurrent models and TCN that use the MLP decoder. Allows user to specify the activation function that was previously hardcoded to ReLU. 

Adds a warning when the parameter is set but unusable.

Fixes #1608

### DIFF
--- a/neuralforecast/models/rnn.py
+++ b/neuralforecast/models/rnn.py
@@ -216,6 +216,12 @@ class RNN(BaseModel):
             self.proj = nn.Linear(
                 self.encoder_hidden_size, self.loss.outputsize_multiplier
             )
+            if self.decoder_activation != "ReLU":
+                warnings.warn(
+                    "decoder_activation is ignored when recurrent=True, since the "
+                    "model has no MLP decoder: the output is a linear projection of "
+                    "the recurrent state."
+                )
         else:
             self.mlp_decoder = MLP(
                 in_features=self.encoder_hidden_size + self.futr_exog_size,
--- a/neuralforecast/models/rnn.py
+++ b/neuralforecast/models/rnn.py
@@ -36,6 +36,7 @@ class RNN(BaseModel):
         context_size (deprecated).
         decoder_hidden_size (int): size of hidden layer for the MLP decoder.
         decoder_layers (int): number of layers for the MLP decoder.
+        decoder_activation (str): activation function for the MLP decoder, see [activations collection](https://docs.pytorch.org/docs/stable/nn.html#non-linear-activations-weighted-sum-nonlinearity). Unused when `recurrent=True`, since no MLP decoder is created.
         futr_exog_list (str list): future exogenous columns.
         hist_exog_list (str list): historic exogenous columns.
         stat_exog_list (str list): static exogenous columns.
@@ -95,6 +96,7 @@ class RNN(BaseModel):
         context_size: Optional[int] = None,
         decoder_hidden_size: int = 128,
         decoder_layers: int = 2,
+        decoder_activation: str = "ReLU",
         futr_exog_list=None,
         hist_exog_list=None,
         stat_exog_list=None,
@@ -190,6 +192,7 @@ class RNN(BaseModel):
         # MLP decoder
         self.decoder_hidden_size = decoder_hidden_size
         self.decoder_layers = decoder_layers
+        self.decoder_activation = decoder_activation
 
         # RNN input size (1 for target variable y)
         input_encoder = (
@@ -219,7 +222,7 @@ class RNN(BaseModel):
                 out_features=self.loss.outputsize_multiplier,
                 hidden_size=self.decoder_hidden_size,
                 num_layers=self.decoder_layers,
-                activation="ReLU",
+                activation=self.decoder_activation,
                 dropout=0.0,
             )
             if self.h > self.input_size:
--- a/neuralforecast/models/tcn.py
+++ b/neuralforecast/models/tcn.py
@@ -32,6 +32,7 @@ class TCN(BaseModel):
         context_size (int): size of context vector for each timestamp on the forecasting window.
         decoder_hidden_size (int): size of hidden layer for the MLP decoder.
         decoder_layers (int): number of layers for the MLP decoder.
+        decoder_activation (str): activation function for the MLP decoder, see [activations collection](https://docs.pytorch.org/docs/stable/nn.html#non-linear-activations-weighted-sum-nonlinearity).
         futr_exog_list (str list): future exogenous columns.
         hist_exog_list (str list): historic exogenous columns.
         stat_exog_list (str list): static exogenous columns.
@@ -87,6 +88,7 @@ class TCN(BaseModel):
         context_size: int = 10,
         decoder_hidden_size: int = 128,
         decoder_layers: int = 2,
+        decoder_activation: str = "ReLU",
         futr_exog_list=None,
         hist_exog_list=None,
         stat_exog_list=None,
@@ -169,6 +171,7 @@ class TCN(BaseModel):
         # MLP decoder
         self.decoder_hidden_size = decoder_hidden_size
         self.decoder_layers = decoder_layers
+        self.decoder_activation = decoder_activation
 
         # TCN input size (1 for target variable y)
         input_encoder = (
@@ -194,7 +197,7 @@ class TCN(BaseModel):
             out_features=self.loss.outputsize_multiplier,
             hidden_size=self.decoder_hidden_size,
             num_layers=self.decoder_layers,
-            activation="ReLU",
+            activation=self.decoder_activation,
             dropout=0.0,
         )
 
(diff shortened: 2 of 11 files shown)


neuralforecast/models/rnn.py
```diff
@@ -213,13 +216,19 @@ def __init__(
             self.proj = nn.Linear(
                 self.encoder_hidden_size, self.loss.outputsize_multiplier
             )
+            if self.decoder_activation != "ReLU":
```
ngupta23: This warning is missing in TCN and DilatedRNN - Is this intentional?

neuralforecast/models/tcn.py
```diff
@@ -87,6 +88,7 @@ def __init__(
         context_size: int = 10,
         decoder_hidden_size: int = 128,
         decoder_layers: int = 2,
+        decoder_activation: str = "ReLU",
```
ngupta23: Would it be worth making this `decoder_activation: Optional[str] = None`, and then warn on `is not None`, and pass decoder_activation or "ReLU" into MLP? That way you can (1) really differentiate if user is passing a value, (2) this guards against change in the default values in the future.

neuralforecast/models/rnn.py
```diff
@@ -36,6 +36,7 @@ class RNN(BaseModel):
         context_size (deprecated).
         decoder_hidden_size (int): size of hidden layer for the MLP decoder.
         decoder_layers (int): number of layers for the MLP decoder.
+        decoder_activation (str): activation function for the MLP decoder, see [activations collection](https://docs.pytorch.org/docs/stable/nn.html#non-linear-activations-weighted-sum-nonlinearity). Unused when `recurrent=True`, since no MLP decoder is created.
```
ngupta23: Docstrings link the full PyTorch activations page, but `MLP` only accepts the names in `ACTIVATIONS` (`_modules.py:17`) — and validates only when `num_layers >= 2`. So `GRU(recurrent=True, decoder_activation="ReLU6")` constructs fine and warns "ignored", while the same string with `recurrent=False` raises a bare `AssertionError` from inside `MLP`. Worth pointing at `ACTIVATIONS` instead. Note this wording is inherited from `xlstm.py:42`, so it's a pre-existing convention, not something this PR introduced so may be fine to leave as is.

tests/test_models/test_deepar.py
ngupta23, line 1: No tests for TCN or DilatedRNN

tests/test_models/test_deepar.py
```diff
@@ -34,3 +36,20 @@ def my_config_new(trial):
     my_config['lstm_hidden_size'] = 8
     model = AutoDeepAR(h=12, config=my_config, backend='ray', num_samples=1, ray_options=RayOptions(cpus=1))
     model.fit(dataset=dataset)
+
+
+def test_deepar_decoder_activation_ignored_without_hidden_layers():
```
ngupta23: The tests never assert the activation is actually applied. All four new tests only check whether a warning fires. The strongest positive case is:

```python
  assert_no_decoder_activation_warning(
      DeepAR, **kwargs, decoder_hidden_layers=2, decoder_hidden_size=8,
      decoder_activation="Tanh",
  )
```

ngupta23:
The DeepAR `decoder_hidden_size=0 trap. DeepAR(h=4, input_size=8, decoder_hidden_layers=2, decoder_activation="Tanh")` with the default `decoder_hidden_size=0` builds `Linear(0, 0)` hidden layers - a decoder that trains to a constant. The new docstring at `deepar.py:30` steers users toward `decoder_hidden_layers>0` to make the activation take effect, so this PR makes it easier to hit; the PR's own test sidesteps it by passing  `decoder_hidden_size=8`.

Pre-existing, so a follow-up issue rather than a blocker, but worth recording while it's in view.

JQGoh (dismissed):

ngupta23 (commented):
Thanks for the rework — `resolve_decoder_activation` is a better answer than what I suggested. Collapsing six hand-rolled warning blocks into one helper, with `ignored_because` doubling as the warning text, means the message can't drift from the condition. All three of my earlier comments are addressed, and the new `assert_decoder_activation_applied` helper closes the gap where the tests only checked for warnings rather than that the activation actually reached the decoder.

I built the branch and ran the six affected test files — 18 passed. Two things I'd still like to sort out, left inline, plus one that's outside the diff:

**xLSTM wasn't brought along.** It still has `decoder_activation: str = "GELU"` (`xlstm.py:106`), passes it straight to `MLP` with no `resolve_decoder_activation`, no ignore-warning when `decoder_layers=1`, and the old PyTorch-page docstring at `xlstm.py:42`. After this PR it's the one model of seven with the old shape. The `default=` parameter on the new helper covers the `"GELU"` default, so it should be a small change — happy for it to be a follow-up if you'd rather keep this PR scoped.

Opened #1615 for the pre-existing DeepAR `decoder_hidden_size=0` trap, which is adjacent to this but not yours to fix here.

neuralforecast/common/_modules.py
```diff
@@ -40,6 +41,24 @@
...
+        ignored_because (Optional[str]): reason the decoder cannot apply an
+            activation, or None when it can. Used to warn a user who set
+            `decoder_activation` in a configuration where it has no effect.
+        default (str): activation used when the user did not set one.
+
+    Returns:
+        (str): the activation name to hand to `MLP`.
+    """
+    if activation is not None and ignored_because is not None:
+        warnings.warn(f"decoder_activation is ignored when {ignored_because}.")
```
ngupta23: `resolve_decoder_activation` doesn't validate against `ACTIVATIONS`, so the same typo behaves differently depending on which branch swallows it. Verified on this commit:

```python
GRU(h=4, input_size=8, recurrent=True, decoder_activation="NotAnActivation")
# constructs fine; self.decoder_activation == 'NotAnActivation', and that string
# goes into hparams

TCN(h=4, input_size=8, decoder_layers=1, decoder_activation="NotAnActivation")
# warns "ignored", then AssertionError from MLP
```

The recurrent path never builds an `MLP`, so nothing checks it. The docstrings now promise "one of `ACTIVATIONS`", but that path doesn't enforce it.

Adding the assert here would close it uniformly and let you drop the one in `MLP.__init__` back to where it was:

```python
if activation is not None:
    assert activation in ACTIVATIONS, f"{activation} is not in {ACTIVATIONS}"
```

neuralforecast/common/_modules.py
```diff
@@ -79,13 +98,14 @@ def __init__(
     ):
         super().__init__()
 
+        assert activation in ACTIVATIONS, f"{activation} is not in {ACTIVATIONS}"
```
ngupta23: Hoisting this above the `num_layers == 1` branch is the right direction, but note it reaches past the models in this PR: every `MLP` caller now validates even when the activation is unused, `xLSTM` included. Any existing config with `decoder_layers=1` and an out-of-list name that silently worked before will now raise.

I think that's the correct behavior, just worth a release-note line so it doesn't surprise anyone on upgrade. (If you take the suggestion on `resolve_decoder_activation` above, this can move back inside the `else`.)

ngupta23 (approved):
Approving. Verified on `4a4c0e84`: the six affected test files pass (18 tests), `ruff check neuralforecast/` is clean, and I confirmed the two open points directly.

Validation is now uniform — a typo raises the same way regardless of which branch would have swallowed it:

```
GRU(recurrent=True,     decoder_activation="NotAnActivation")  → AssertionError
TCN(decoder_layers=1,   decoder_activation="NotAnActivation")  → AssertionError
GRU(decoder_layers=2,   decoder_activation="NotAnActivation")  → AssertionError
DeepAR(hidden_layers=0, decoder_activation="NotAnActivation")  → AssertionError
```

while a valid-but-ignored value still warns rather than raises, and the assert fires before the warning so a typo doesn't get a misleading "ignored" message first.

Moving the assert back inside `MLP`'s `else` branch was the better call over what I suggested — `MLP` is now byte-identical to `main`, so nothing outside this PR's models changes behavior. **Disregard my earlier note about needing a release-note line**; there's no upgrade-visible change any more.

Agreed on leaving xLSTM as is. Beyond the default being deliberate, `decoder_activation` was already public there before this PR, so switching it to `Optional[str] = None` would change an existing signature and its hparams round-trip for current users — a cost the other six don't carry, since the parameter is new in those. I've filed #1616 for the narrower docstring/ignore-warning gap in xLSTM, which is pre-existing and identical to `main`; nothing for this PR.

Also still open from earlier: #1615 for the DeepAR `decoder_hidden_size=0` trap. Also pre-existing, also not for this PR.

PR #1618, opened 2026-09-16
[FIX] Validate and warn decoder_activation when decoder_layers=1

### LINKED ISSUE #1616: xLSTM: decoder_activation is neither validated nor warned about when decoder_layers=1
`xLSTM` is the one model with an MLP decoder that doesn't validate `decoder_activation` or warn when it can't be applied. Verified on `main` (309578e1):

```python
from neuralforecast.models import xLSTM

k = dict(h=4, input_size=8, max_steps=1)

xLSTM(**k, decoder_layers=1, decoder_activation="Tanh")             # silently dropped, no warning
xLSTM(**k, decoder_layers=1, decoder_activation="NotAnActivation")  # constructs fine
xLSTM(**k, decoder_layers=2, decoder_activation="NotAnActivation")  # AssertionError
```

`MLP` skips its `ACTIVATIONS` assert when `num_layers == 1` (`_modules.py:110`), so with `decoder_layers=1` the decoder collapses to a bare `nn.Linear` and anything passed for `decoder_activation` is discarded unchecked.

#1612 added `resolve_decoder_activation` (`_modules.py:44`) and routed DeepAR, DilatedRNN, GRU, LSTM, RNN and TCN through it, which validates the name up front and warns when the decoder can't consume it. `xLSTM` still passes `decoder_activation` straight to `MLP` at `xlstm.py:228`.

Deliberately left out of #1612, and I agree with that call: the `"GELU"` default is an architectural choice, and unlike the other six, `decoder_activation` was already public on `xLSTM`, so changing it to `Optional[str] = None` would alter an existing signature and its hparams round-trip.

The narrower fix needs no signature change — keep `decoder_activation: str = "GELU"` and pass it through the helper:

```python
decoder_activation = resolve_decoder_activation(
    decoder_activation,
    "decoder_layers=1, since the decoder is then a single linear layer"
    if decoder_layers == 1
    else None,
    default="GELU",
)
```

The helper only asserts when the value is not None, so a `str`-typed parameter with a non-None default is always validated — which is the desired behavior here.

Also worth updating the docstring at `xlstm.py:42`: it links the full PyTorch non-linear-activations page, but only the names in `ACTIVATIONS` (`_modules.py:17`) are accepted. The other six models were changed in #1612 to point at `ACTIVATIONS` instead.

Pre-existing behavior, unchanged by #1612.

marcopeix:
Same behaviour as other recurrent models and TCN. Add validation and warnings for decoder_activation when decoder_layers=1.

Fixes #1616


tests/test_models/test_xlstm.py
```diff
@@ -0,0 +1,30 @@
+import pytest
```
ngupta23, line 1: Consider adding `pytest.importorskip("xlstm")` at the top of this module.

`xlstm` is not in `project.dependencies` — it is only in the `dev` dependency group — and `xLSTM.__init__` raises `ImportError("Please install `xlstm`...")` when it is absent. This is the first test under `tests/` that instantiates `xLSTM`, so in any environment synced without the dev group, all six constructions in this test error out instead of skipping.

CI itself is unaffected (it runs `uv sync --group dev`), so this is about contributors with a lighter env. The repo convention for optional deps is `pytest.importorskip` — see `tests/test_categorical.py:624` (captum), `tests/test_categorical.py:796` (pyspark/fugue), and `tests/test_core.py:1470` (torch.distributed).

neuralforecast/models/xlstm.py
```diff
@@ -219,6 +219,14 @@ def __init__(
             )
         self.hist_encoder = xLSTMBlockStack(block_stack_config)
 
+        decoder_activation = resolve_decoder_activation(
```
ngupta23: Suggest moving this `resolve_decoder_activation` call above the `# Architecture` section, so it runs before `xLSTMBlockStack` is built.

Every sibling model resolves the activation *before* constructing the encoder — `tcn.py:174`, `dilated_rnn.py:444`, `rnn.py:203`, `lstm.py:197`, `gru.py:211`. Here it lands after `self.hist_encoder = xLSTMBlockStack(block_stack_config)`, which has two consequences:

1. `xLSTM(h=4, input_size=8, encoder_hidden_size=1024, encoder_n_blocks=8, decoder_activation="Taanh")` allocates and initializes the entire block stack before raising the `AssertionError` on the typo.
2. It weakens the helper's own stated intent — "a typo is rejected even when the decoder is never built" (`_modules.py:47`). Without `xlstm` installed, the typo is masked entirely by the `ImportError` raised earlier in `__init__`.

Hoisting the call fixes both and matches the established placement.

tests/test_models/test_xlstm.py
```diff
@@ -0,0 +1,32 @@
...
+
+from neuralforecast.models import xLSTM
+
+from .test_helpers import (
+    assert_decoder_activation_applied,
+    assert_no_decoder_activation_warning,
+)
+
+
+def test_xlstm_decoder_activation():
```
ngupta23, line 14: Non-blocking: this file omits the `check_model` / `check_args` blocks that every sibling `tests/test_models/test_<model>.py` carries — see `test_tcn.py:15` (`check_model(TCN, ["airpassengers"])`) and `test_tcn.py:19`/`:23` (`check_args(AutoTCN, exclude_args=['cls_model'])`).

The consequence is that `AutoxLSTM` still has no `check_args` coverage, so a future `BaseAuto` signature change would break it silently. The `check_args` call alone would close that and needs no training, so it's cheap to add here if you'd like — otherwise happy to see it as a follow-up.

ngupta23 (approved):
Both review comments are addressed in 9a7354dd — thanks. Verified on this head:

- **`importorskip`**: blocked `xlstm` via a meta-path hook and ran this file at both commits — `1 failed` at 07d032dd, `1 skipped` now (`could not import 'xlstm'`).
- **Hoisted validation**: `xLSTM(h=4, input_size=8, encoder_hidden_size=1024, encoder_n_blocks=8, decoder_activation="Taanh")` now raises `AssertionError` in 0.003s with no block stack allocated.
- `ruff check` clean on both files; `tests/test_models/test_xlstm.py` passes normally (`1 passed`).

On the secondary point in my earlier comment — that a typo is masked when `xlstm` isn't installed, since the resolve call still runs after the `ImportError` — leaving that as-is is the right call. The model can't be constructed at all without `xlstm`, so validating first would only change which error the user sees.

One non-blocking follow-up inline. Separately, and definitely out of scope here: `ACTIVATIONS` contains `Threshold` and `GLU`, neither of which can actually work in `MLP` (`Threshold` raises `TypeError: Threshold.__init__() missing 2 required positional arguments`; `GLU` halves the hidden dim into a shape mismatch at forward). That's pre-existing from #1612 and affects all 7 models that take `decoder_activation` — worth its own PR to prune the list.

PR #1620, opened 2026-09-17
[FIX] Floor pytorch lightning to 2.6.6

### LINKED ISSUE #1619: Relax pytorch-lightning <2.6.0 cap: blocks the CVE-2026-58659 fix in 2.6.6
## Summary

`pyproject.toml` caps `pytorch-lightning` at `<2.6.0`. That cap was added in #1513 as a reaction to the supply-chain compromise GHSA-w37p-236h-pfx3, which affects **only 2.6.2 and 2.6.3** — both since quarantined from PyPI.

The cap is now doing two unintended things:

1. It excludes **2.6.1**, the exact version Lightning's own advisory tells users to pin to.
2. It blocks **2.6.6**, which contains the only upstream fixes for a `load_from_checkpoint` RCE (CVE-2026-58659) that **bypasses the `weights_only=True` mitigation**.

Net effect: a constraint added for security reasons now holds users on a branch that cannot receive a security fix.

## Current state

https://github.com/Nixtla/neuralforecast/blob/main/pyproject.toml (v3.2.2):

```toml
dependencies = [
    ...
    "pytorch-lightning>=2.0.0,<2.6.0",
    ...
]
```

Introduced in #1513 (merged 6 May 2026), commit `690317f`:

```diff
-    "pytorch-lightning>=2.0.0",
+    "pytorch-lightning>=2.0.0,<2.6.0",
```

with the rationale:

> Recent pytorch-lightning release had a security breach, to fix that we're missing the dependency
> Read more about it here: GHSA-w37p-236h-pfx3

That was the right call at the time. The situation has since changed.

## Why the cap is now broader than the threat

**GHSA-w37p-236h-pfx3 / CVE-2026-44484** — https://github.com/advisories/GHSA-w37p-236h-pfx3

- Affected versions: `= 2.6.2`, `= 2.6.3`
- Lightning's stated remediation: *"Pin PyTorch Lightning to version `2.6.1`"*
- Lightning's stated action: *"Quarantined malicious versions from PyPI"*

So 2.6.2 and 2.6.3 are no longer resolvable from PyPI at all, and 2.6.1 — the recommended version — is excluded by `<2.6.0`. The cap currently protects against nothing that a resolver could still select.

2.6.4 (20 May), 2.6.5 (27 May) and 2.6.6 (10 Sep) all post-date the incident and are outside the advisory's affected set.

## Why this now matters for security

**CVE-2026-58659** — https://github.com/advisories/GHSA-qqmf-gpg7-g8gw

> PyTorch Lightning through 2.6.5, fixed in commit `d710d68`, contains a remote code execution vulnerability in the `_load_state` function that imports and executes attacker-controlled module names from checkpoint `_instantiator` hyperparameters. Attackers can craft malicious checkpoint files that **bypass `weights_only=True` protections** to execute arbitrary code when `LightningModule.load_from_checkpoint` is called.

CVSS v4 8.4 High, CWE-470. Upstream issue Lightning-AI/pytorch-lightning#21822, fix PR #21832.

Fixed in **2.6.6** — https://github.com/Lightning-AI/pytorch-lightning/releases/tag/2.6.6:

> - Fixed arbitrary code execution in `load_from_checkpoint` by restricting the `_instantiator` hyperparameter to an allowlist of trusted instantiators (#21832)
> - Fixed arbitrary code execution in `load_from_checkpoint` by rejecting a checkpoint `_class_path` that does not resolve to an already imported subclass of the loaded class (#21914)

This is directly relevant to neuralforecast: `NeuralForecast.load()` and `BaseModel.load()` deserialize checkpoints, so any downstream user loading a checkpoint they did not produce is on the vulnerable path, and the usual `weights_only=True` advice does not help here.

Worth flagging: this advisory is filed against the **`lightning`** PyPI package, not `pytorch-lightning`, with a range of `< 2022.6.15` that appears to be a CalVer sorting artifact. Scanners keyed on `pytorch-lightning` will not surface it, even though the prose says "through 2.6.5". That makes it easy to miss.

## Proposed change

```toml
"pytorch-lightning>=2.6.6",
```

or, if a wider floor is preferred for compatibility:

```toml
"pytorch-lightning>=2.0.0,!=2.6.2,!=2.6.3",
```

The first is the stronger option, since it is the only one that guarantees users get the CVE-2026-58659 fix. The second preserves the existing support range while still excluding the compromised releases explicitly.

## On compatibility

I reviewed the 2.6.x release notes and did not find a breaking change that looks likely to affect neuralforecast. The notable ones are `weights_only` exposed on `Trainer.{fit,validate,test,predict}`, Rich progress bar becoming the default when `rich` is installed, Python 3.9 support dropped in 2.6.1, `to_torchscript` deprecated in 2.6.1, and the Neptune logger removed in 2.6.4.

I have not run the neuralforecast test suite against 2.6.6, so I can't claim this is a drop-in change — that is worth verifying in CI. I'm raising it because the constraint's original justification no longer holds, not because I've validated the upgrade.

## Context

Found while remediating dependency CVEs in a downstream project that uses `neuralforecast.models.TFT`. The cap is what prevents us from resolving a patched `pytorch-lightning`, and I suspect other users pinning neuralforecast are in the same position without realising the advisory is filed under a different package name.

Happy to open a PR if the maintainers agree on which of the two constraint forms is preferred.

marcopeix:
Closes #1619

The `<2.6.0` cap was added in #1513 to dodge the compromised 2.6.2/2.6.3 releases (GHSA-w37p-236h-pfx3). Those versions are gone from PyPI, so it pins users to versions affected by CVE-2026-58659, an RCE in `load_from_checkpoint` that bypasses `weights_only=True` and hits anyone calling `NeuralForecast.load()`. Correct fix is in 2.6.6.


janrth (approved):
LGTM

PR #1621, opened 2026-09-19
Feat/deepar hist exog 20260919

immanuelk1m:
Closing this submission because its branch includes unrelated fork history (99 commits / 156 changed files) and conflicts with upstream main. Following CONTRIBUTING.md's single-purpose PR and resubmission guidance, the DeepAR historical-exogenous change will be ported to a clean branch based on upstream main, tested there, and submitted separately. The original implementation is in immanuelk1m/neuralforecast#17. No changes to upstream main are requested by this closed PR.

PR #1623, opened 2026-09-19
[FEAT] Support historical exogenous variables in DeepAR

### LINKED ISSUE #1622: [Models] Support historical-only exogenous variables in DeepAR
### Description

Support `DeepAR(hist_exog_list=...)` for covariates observed only before the forecast origin, alongside the existing known-future and static inputs.

On upstream `main` at `344aaffd504245ff661bd9e211f220f9214a1876`, the constructor exposes `hist_exog_list`, but `EXOGENOUS_HIST` is `False` and the LSTM input contains only the target, future exogenous variables, and static variables. For example:

```python
from neuralforecast.models import DeepAR

model = DeepAR(h=2, input_size=52, hist_exog_list=["inventory", "dxy"])
```

The requested behavior is to use historical covariates to condition the recurrent state, without requiring their future observations at prediction time. Known-future covariates should retain their existing `futr_exog_list` semantics. The existing recurrent `BaseModel` path already handles historical-input alignment and masks the forecast range.

This is a past-only extension to the covariate-conditioned architecture in [DeepAR, arXiv:1704.04110v3, Sections 3 and 3.4](https://arxiv.org/abs/1704.04110v3). The paper itself assumes covariates are available throughout the prediction range.

### Use case

For commodity forecasting, observed inventory and exchange-rate histories may be available at the forecast origin while their future values are unknown. I would like to evaluate DeepAR with the same historical-only information set used by other NeuralForecast models, without supplying future realized covariates or constructing a separate forecast for each covariate.

Historical data still need to respect publication delays and revisions. This request concerns the model input interface and does not claim an accuracy improvement.

I checked the existing DeepAR issues and PRs. The earlier submission #1621 has been closed because its branch included unrelated fork changes; a focused implementation based directly on upstream main is being validated for a replacement PR.

immanuelk1m:
Closes #1622. Supersedes #1621.

## Problem

`DeepAR` exposes `hist_exog_list` in its constructor but rejects nonempty lists because `EXOGENOUS_HIST` is `False`. Its LSTM input also omits the historical stream. This prevents using observed covariate histories when their future values are unknown, such as inventory or exchange-rate histories in commodity forecasting.

## Changes

- Enable `EXOGENOUS_HIST`, add `hist_exog_size` to the LSTM input width, and concatenate the historical stream in `forward`.
- Reuse the existing recurrent `BaseModel` alignment, forecast-range masking, normalization, and categorical embeddings. Observed historical covariates condition the recurrent state; their future observations are not required.
- Preserve the no-history computation, including direct `forward` callers that omit the unused `hist_exog` key. Keep known-future/static inputs and the current upstream decoder activation/width validation behavior.
- Document the historical-only semantics in the model docstring and add focused regression tests to the existing DeepAR test file.

This branch contains **one commit directly on upstream `344aaffd504245ff661bd9e211f220f9214a1876`, changing only two files**: `neuralforecast/models/deepar.py` and `tests/test_models/test_deepar.py`. There are no changes to shared preprocessing, dependencies, workflows, or other models. The unrelated fork history from #1621 is excluded.

## Relationship to the paper

[DeepAR, arXiv:1704.04110v3](https://arxiv.org/abs/1704.04110v3), Section 3 / Equation (1), conditions the recurrent network on the previous target and covariates. Sections 3 and 3.4 explicitly assume covariates are available throughout the prediction range. **Historical-only inputs are an extension of that formulation.** `futr_exog_list` retains the known-future behavior. This PR leaves the existing likelihood, recurrent feedback, and sampling implementation unchanged.

## Validation

[Successful validation run](https://github.com/immanuelk1m/neuralforecast/actions/runs/35442379773) checks out the exact proposed commit `ebea9f3b958760ff97c948fa5bcd7ac482734e3a`. The validation workflow lives on a separate fork-only branch and is excluded from this PR.

Both Ubuntu CPU jobs, **Python 3.11.16 and 3.12.3**, passed with PyTorch `2.14.0+cpu`, PyTorch Lightning `2.6.6`, and pandas `2.3.3`:

| Check | Result in each Python environment |
| --- | --- |
| Repository-configured pre-commit hooks on the two changed files | Ruff and mypy passed; requirements hook had no matching files |
| New historical-input tests with the upstream model restored | All 12 failed with `DeepAR does not support historical exogenous variables`, confirming the regression tests detect the missing feature |
| Historical-input tests with the patch | 14 passed, including the two no-history compatibility cases |
| Existing DeepAR and related recurrent-model tests | 17 passed |
| Patch scope and whitespace | Exactly two changed files; `git diff --check` passed |

Commands used after installing the project and development dependencies:

```sh
pre-commit run --files neuralforecast/models/deepar.py tests/test_models/test_deepar.py
pytest tests/test_models/test_deepar.py -k hist_exog --no-cov --tb=short
pytest tests/test_models/test_deepar.py tests/test_models/test_rnn.py \
  tests/test_models/test_lstm.py tests/test_models/test_gru.py \
  tests/test_models/test_tcn.py tests/test_models/test_dilated_rnn.py \
  -k 'not auto and not hist_exog' --no-cov --tb=short
```

Coverage includes output shapes, historical-input gradients and sensitivity, recurrent state continuity, unchanged no-history outputs, forecast-range history poisoned with large values or NaN under identity/standard scaling, categorical historical inputs, and actual `NeuralForecast.fit` / validation / `predict` without future historical columns for `h_train=1` and `3` with both historical-only and mixed covariates.

Validation used a CPU environment installed with `uv pip`; pandas was pinned only in that environment. Auto/Ray/Optuna tests, the full repository suite and coverage gate, GPU execution, and accuracy benchmarks were not run. No forecast-accuracy improvement is claimed.

## Usage

```python
from neuralforecast.models import DeepAR

model = DeepAR(
    h=2,
    input_size=52,
    hist_exog_list=["inventory", "dxy"],
    scaler_type="standard",
)
nf = NeuralForecast(models=[model], freq="W-MON")
nf.fit(df=train_df)  # unique_id, ds, y, inventory, dxy
predictions = nf.predict()  # No future inventory or DXY observations required.
```

Column names are illustrative. Historical inputs must respect publication delays and revisions. Adding `futr_exog_list` still requires those known-future columns over the prediction horizon.


neuralforecast/models/deepar.py
```diff
@@ -208,10 +217,14 @@ def forward(self, windows_batch):
 
         # Parse windows_batch
         encoder_input = windows_batch["insample_y"]  # <- [B, T, 1]
+        hist_exog = windows_batch.get("hist_exog")
```
marcopeix: Use `windows_batch["hist_exog"]` please

marcopeix (changes requested):
Thanks @immanuelk1m for the PR! See my comments below. Also, we should update the capabilities notebook to show that DeepAR supports historical exog.

marcopeix (approved):
Thank you for your work!

